PR #13, opened 2020-09-02
Add github action to build project

rzgry:


rzgry:
#14 should be merged instead.

PR #27, opened 2020-09-21
add windows CI tests

kathrynkodama:
Add CI tests to run on Windows

Signed-off-by: Kathryn Kodama <kathryn.s.kodama@gmail.com>


rzgry (commented):
If the steps for ubuntu and windows are the same i think you can use a OS matrix. https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-running-with-more-than-one-operating-system

rzgry (approved):

PR #42, opened 2021-02-01
Liberty version and installed features fallback

### LINKED ISSUE #20: Detect version of liberty being used
Detect version of liberty being used to suggest appropriate list of features, completion options etc.

kathrynkodama:
Fixes #20 

Added the following functionality:
1) Determine the Liberty version from the `openliberty.properties` file found in the workspace.
When a new properties file is detected, we watch this file in a separate thread. If this file is modified or the enclosing folder is deleted, we treat this as the Liberty installation being modified. The associated `isLibertyInstalled` value (see workspace folders below) is set to `false`.

2) Track workspace folders
See the `LibertyWorkspace.java` class. Each workspace folder has an associated URI, Liberty version, and `isLibertyInstalled` value. The Liberty version is used as the version cache. The `isLibertyInstalled` value is used to determine whether the LS should look for a new `openliberty.properties` file.

3) If the features list cannot be retrieved from Maven, fallback to a list of installed features.
Uses the `ws-featureslist.jar` to generate the list of installed features. Generates a featurelist.xml document into the LemMinx cache. Then uses the existing `Feature.java` class and the new `FeatureInfo.java` class to parse the xml document into a feature list. Only the `name`, `description`, and `shortDescription` are set for the installed features.
image

### DIFF
--- a/lemminx-liberty/src/main/java/io/openliberty/lemminx/liberty/LibertyCompletionParticipant.java
+++ b/lemminx-liberty/src/main/java/io/openliberty/lemminx/liberty/LibertyCompletionParticipant.java
@@ -30,7 +30,9 @@ public class LibertyCompletionParticipant extends CompletionParticipantAdapter {
     public void onXMLContent(ICompletionRequest request, ICompletionResponse response)
             throws IOException, BadLocationException {
         if (!LibertyUtils.isServerXMLFile(request.getXMLDocument()))
-            return;
+            return;    
+
+        LibertyUtils.getVersion(request.getXMLDocument());
 
         DOMElement parentElement = request.getParentElement();
         if (parentElement == null || parentElement.getTagName() == null)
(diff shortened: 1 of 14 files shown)


lemminx-liberty/src/main/java/io/openliberty/lemminx/liberty/LibertyCompletionParticipant.java
```diff
@@ -69,16 +71,22 @@ private CompletionItem buildFeatureCompletionItem(Feature feature, DOMElement fe
...
-    private List<CompletionItem> buildCompletionItems(DOMElement featureElement, DOMDocument document,
+    private List<CompletionItem> buildCompletionItems(DOMElement featureElement, DOMDocument domDocument,
             List<String> existingFeatures) {
-        final String libertyVersion = SettingsService.getInstance().getLibertyVersion();
+
+        String libertyVersion = LibertyUtils.getVersion(domDocument);
+        if (libertyVersion == null) {
+            // try to get version from installed Liberty
+            libertyVersion = LibertyUtils.getVersion(domDocument);
+        }
```
rzgry, line 81: Seems to be calling `LibertyUtils.getVersion(domDocument);` twice? Is this on purpose? 

other files seem to do this instead
```
        String libertyVersion = SettingsService.getInstance().getLibertyVersion();
        if (libertyVersion == null) {
            // try to get version from installed Liberty
            libertyVersion = LibertyUtils.getVersion(domDocument);
        }
```

lemminx-liberty/src/main/java/io/openliberty/lemminx/liberty/services/FeatureService.java
```diff
@@ -59,13 +70,14 @@ private FeatureService() {
     // Only need the public features
     ArrayList<Feature> publicFeatures = readPublicFeatures(reader);
 
+    LOGGER.info("returning public features: " + publicFeatures.size());
```
rzgry, line 73: Maybe make this more descriptive? something like
"Returning public features from maven"

lemminx-liberty/src/main/java/io/openliberty/lemminx/liberty/services/FeatureService.java
```diff
@@ -96,23 +109,24 @@ private FeatureService() {
...
-            .filter(f -> f.getWlpInformation().getVisibility().equals(LibertyConstants.PUBLIC_VISIBILITY))
-            .forEach(publicFeatures::add);
+        .filter(f -> f.getWlpInformation().getVisibility().equals(LibertyConstants.PUBLIC_VISIBILITY))
+        .forEach(publicFeatures::add);
     defaultFeatureList = publicFeatures;
     return publicFeatures;
   }
 
-  public List<Feature> getFeatures(String libertyVersion, int requestDelay) {
+  public List<Feature> getFeatures(String libertyVersion, int requestDelay, String documentURI) {
```
rzgry, line 118: Maybe add a comment above on how this function works.

So tries to fetch the list of all possible features from maven for the specific liberty version. If that fails, tries to determine the installed features. Else fallback to a hardcoded default list.

lemminx-liberty/src/main/java/io/openliberty/lemminx/liberty/services/FeatureService.java
```diff
@@ -121,18 +135,77 @@ private FeatureService() {
...
+        File tempFeaturesList = File.createTempFile("featureslist", ".xml", tempDir);
+        String[] cmd = { "java", "-jar", featureListJAR.toAbsolutePath().toString(),
+            tempFeaturesList.getAbsolutePath() };
+
+        Process proc = Runtime.getRuntime().exec(cmd);
+        BufferedReader in = new BufferedReader(new InputStreamReader(proc.getInputStream()));
+        String s = null;
+        while ((s = in.readLine()) != null) {
+          LOGGER.info(s);
+        }
```
rzgry, line 187: I think this can be removed?

rzgry (approved):

PR #54, opened 2022-04-15
Add launch.json for attaching java debugger

kevinwlau:
To debug, have both `open-liberty-tools-vscode` and `liberty-language-server` projects at the root of VS Code workspace.
image

Run `npm run build` in `open-liberty-tools-vscode`

Run `Run Extension` first, a new VS Code window will open with the extension running.
Then run `Debug Attach to lemminx`
image

Set your breakpoints and use the new VS Code window to trigger actions (ie. hover or completion)

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -7,7 +7,7 @@
 		{
 			"type": "java",
 			"request": "attach",
-			"name": "Debug Attach to lemminx",
+			"name": "Debug attach liberty-lemminx",
 			"hostName": "localhost",
 			"projectName": "liberty-langserver-lemminx",
 			"port": 1054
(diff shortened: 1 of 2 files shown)


cherylking:
Should we add these instructions to the readme? 

For the step "Then run `Debug Attach to lemminx`", is that from the first VS code window or the second one with the extension running?

kevinwlau:
I'll create a section in the readme on debugging

kevinwlau:
Turns out the `lemminx-liberty` project already had a launch.json configuration for debugging, which explains the above if you didn't have these new changes.

I'll keep this top-level launch.json to consolidate debug configs for lemminx and language server, and to make it easier to setup the whole project.

Issue #40, opened 2021-01-25
Improve logic in determining a config file


kathrynkodama:
Currently the language server is activated for files named `server.xml`. 

A server.xml can point to other xml files using the `include` tag:
https://www.ibm.com/support/knowledgecenter/SSEQTP_liberty/com.ibm.websphere.wlp.doc/ae/twlp_setup_includes.html

We should improve the logic in how we determine which xml files to activate the language server on.

cherylking:
This should also include any files located in pre-defined config dirs including:

`/usr/shared/config`
`/configDropins/override` under the server dir
`/configDropins/defaults` under the server dir

PR #55, opened 2022-04-20
Enable Lang Server for more Config Files

dshimo:
For #40:

`isServerXMLFile` expanded into `isConfigXMLFile`, where now, a cache tracks files found in `configDropins` and files found in `server.xml`'s `include` tags.

The cache is first built when a `LibertyWorkspace` is initialized upon startup. Changes made to known config files will trigger Diagnostics, which provide warning when the pointed file is not found.

Diagnostics Warning Example:
image

As example, `missing.xml` has hover features from being pointed to by the `server.xml` above.
image

### DIFF
--- a/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
+++ b/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
@@ -61,7 +61,7 @@ public class LibertyDiagnosticParticipant implements IDiagnosticsParticipant {
         
     }
 
-    private void validateFeature(DOMDocument domDocument, List<Diagnostic> list, DOMNode node) {
+    private void validateFeature(DOMDocument domDocument, List<Diagnostic> list, DOMNode featureManager) {
         String libertyVersion =  LibertyUtils.getVersion(domDocument);
 
         final int requestDelay = SettingsService.getInstance().getRequestDelay();
@@ -69,7 +69,7 @@ public class LibertyDiagnosticParticipant implements IDiagnosticsParticipant {
         // Search for duplicate features
         // or features that do not exist
         Set<String> includedFeatures = new HashSet<>();
-        List<DOMNode> features = node.getChildren();
+        List<DOMNode> features = featureManager.getChildren();
         for (DOMNode featureNode : features) {
             DOMNode featureTextNode = (DOMNode) featureNode.getChildNodes().item(0);
             // skip nodes that do not have any text value (ie. comments)
--- a/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
+++ b/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
@@ -104,7 +104,7 @@ public class LibertyDiagnosticParticipant implements IDiagnosticsParticipant {
         String docURIString = domDocument.getDocumentURI().replace(File.separator, "/");    // URI requires
         locAttribute = locAttribute.replace(File.separator, "/");
         File configFile = locAttribute.startsWith("./") ? 
-                new File(URI.create(docURIString.substring(0, docURIString.lastIndexOf(File.separator) + 1))
+                new File(URI.create(docURIString.substring(0, docURIString.lastIndexOf("/") + 1))
                         .resolve(locAttribute).normalize()) :
                 new File(locAttribute);
 
--- a/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
+++ b/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
@@ -102,6 +102,7 @@ public class LibertyDiagnosticParticipant implements IDiagnosticsParticipant {
             return;
         }
         String docURIString = domDocument.getDocumentURI();
+        locAttribute = locAttribute.replace("/", File.separator);
         File configFile = locAttribute.startsWith("." + File.separator) ? 
                 new File(URI.create(docURIString.substring(0, docURIString.lastIndexOf(File.separator) + 1)).resolve(locAttribute).normalize()) :
                 new File(locAttribute);
--- a/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
+++ b/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
@@ -101,10 +101,11 @@ public class LibertyDiagnosticParticipant implements IDiagnosticsParticipant {
         if (locAttribute == null) {
             return;
         }
-        String docURIString = domDocument.getDocumentURI();
+        String docURIString = domDocument.getDocumentURI().replace(File.separator, "/");    // URI requires
         locAttribute = locAttribute.replace(File.separator, "/");
-        File configFile = locAttribute.startsWith("." + File.separator) ? 
-                new File(URI.create(docURIString.substring(0, docURIString.lastIndexOf(File.separator) + 1)).resolve(locAttribute).normalize()) :
+        File configFile = locAttribute.startsWith("./") ? 
+                new File(URI.create(docURIString.substring(0, docURIString.lastIndexOf(File.separator) + 1))
+                        .resolve(locAttribute).normalize()) :
                 new File(locAttribute);
 
         DOMNode locNode = node.getAttributeNode("location");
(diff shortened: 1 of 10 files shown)


cherylking:
Two thoughts before I start reviewing:

1) Do we need to also check the `/usr/shared/config/` folder for config files? See this [link](https://www.ibm.com/docs/en/was-liberty/base?topic=liberty-directory-locations-properties) for the directory structure of a Liberty installation.

2) When the diagnostic error is given for a missing config file, can we provide a fix that creates an "empty" config file with the basic `<server>` element in it?

dshimo:
To address the thoughts:
1. That's a simple addition I can throw in. I have an initial check for `configDropins`, but I see now it should instead include `usr/shared/config`, `configDropins/override` and `configDropins/defaults`?
2. From what I understand, Diagnostics can trigger while a user is still typing, so a lot of unnecessary config files could be made.

cherylking:
For question 1, yes I think the three dirs you specified are correct.

For question 2, I was implying that there could be an option for the user to click on `Quick Fix` which would create the file for them in the correct location - not that it would happen automatically. Right now, the diagnostic lists `No quick fixes available`.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
```diff
@@ -22,49 +22,54 @@
...
+    private void validateFeature(DOMDocument domDocument, List<Diagnostic> list, DOMNode node) {
         String libertyVersion =  LibertyUtils.getVersion(domDocument);
 
         final int requestDelay = SettingsService.getInstance().getRequestDelay();
 
         // Search for duplicate features
         // or features that do not exist
         Set<String> includedFeatures = new HashSet<>();
-        List<DOMNode> features = featureManager.getChildren();
+        List<DOMNode> features = node.getChildren();
```
cherylking, line 72: I think it would be clearer to still call the parameter featureManager since that is what is passed in.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
```diff
@@ -90,4 +95,31 @@ private void validateFeatures(DOMDocument domDocument, List<Diagnostic> list) th
...
+        locAttribute = locAttribute.replace(File.separator, "/");
+        File configFile = locAttribute.startsWith("./") ? 
+                new File(URI.create(docURIString.substring(0, docURIString.lastIndexOf("/") + 1))
+                        .resolve(locAttribute).normalize()) :
+                new File(locAttribute);
+
+        DOMNode locNode = node.getAttributeNode("location");
+        Range range = XMLPositionUtility.createRange(locNode.getStart(), locNode.getEnd(), domDocument);
+        try {
+            if (configFile.exists()) {
```
cherylking, line 114: Should we make sure the configFile ends with `.xml`?

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/LibertyWorkspace.java
```diff
@@ -62,8 +82,54 @@ public List<Feature> getInstalledFeatureList() {
...
 
-    public void setInstalledFeatureList(List<Feature> installedFeatureList){
+    public void setInstalledFeatureList(List<Feature> installedFeatureList) {
         this.installedFeatureList = installedFeatureList;
     }
 
+    public void initConfigFileList() {
+        try {
+            List<Path> configPathsList = Files.find(Paths.get(getWorkspaceURI()), Integer.MAX_VALUE, (filePath, fileAttributes) -> 
+                    LibertyUtils.isServerXMLFile(filePath.toString()) || (fileAttributes.isDirectory() && filePath.toString().endsWith("configDropins")))
```
cherylking, line 92: As we discussed, should add `configDropins/overrides`, `configDropins.defaults` and `usr/shared/config` here.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/LibertyWorkspace.java
```diff
@@ -62,8 +82,54 @@ public List<Feature> getInstalledFeatureList() {
...
+        } catch (IOException e) {
+            // workspace URI does not exist
+            e.printStackTrace();
+        }
+    }
+
+    public void scanXMLforInclude(Path filePath) {
+        try {
+            String content = new String(Files.readAllBytes(filePath));
+            Matcher m = Pattern.compile("<include.+location=[\"\'](.+)[\"\']/>").matcher(content);
```
cherylking, line 110: Is this the best way to find `<include>` elements? When using a regex you need to worry about elements that span multiple lines. We could load a DOM Document and find the include elements. If you think regex is the best way to go, double check it and make sure it will handle multiple lines.

cherylking (changes requested):

dshimo:
I made a new issue for implementing the quick fix option, as I think it's a little beyond the scope of this PR and issue.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/LibertyWorkspace.java
```diff
@@ -89,7 +91,7 @@ public void setInstalledFeatureList(List<Feature> installedFeatureList) {
     public void initConfigFileList() {
         try {
             List<Path> configPathsList = Files.find(Paths.get(getWorkspaceURI()), Integer.MAX_VALUE, (filePath, fileAttributes) -> 
-                    LibertyUtils.isServerXMLFile(filePath.toString()) || (fileAttributes.isDirectory() && filePath.toString().endsWith("configDropins")))
+                    LibertyUtils.isServerXMLFile(filePath.toString()) || isConfigDir(filePath, fileAttributes))
```
cherylking, line 94: When the filePath is a config dir, does the dir get added to the list? What about the files within that dir? How do they get added to the list?

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/LibertyWorkspace.java
```diff
@@ -107,7 +109,7 @@ public void initConfigFileList() {
     public void scanXMLforInclude(Path filePath) {
         try {
             String content = new String(Files.readAllBytes(filePath));
-            Matcher m = Pattern.compile("<include.+location=[\"\'](.+)[\"\']/>").matcher(content);
+            Matcher m = Pattern.compile("<include[^<]+location=[\"\'](.+)[\"\']/>").matcher(content);
```
cherylking, line 112: Can you add a comment that explains this regex and what it will match exactly? Explain what is in group 1 since you use that below.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/LibertyWorkspace.java
```diff
@@ -118,6 +120,15 @@ public void scanXMLforInclude(Path filePath) {
         }
     }
 
+    public boolean isConfigDir(Path filePath, BasicFileAttributes fileAttributes) {
+        if (!fileAttributes.isDirectory()) {
+            return false;
+        }
+        return filePath.endsWith(LibertyConstants.WLP_USER_CONFIG_DIR) ||
```
cherylking, line 127: Can filePath be a file that is within one of these dirs, and if so how do they get added to the config file list?

cherylking (commented):
See additional questions.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/LibertyWorkspace.java
dshimo: When filePath is a config dir, it does not get added to the list. All the files within the dir are added with no distinction. Although, that can be changed for thoroughness.
I add them with `Files.list(configPath).forEach(path -> configFiles.add(path.toString()));` in the following `for` loop. 
The dir scan only trigger on initialization.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/LibertyWorkspace.java
dshimo: Group 1 is limited to the contents of what `location` gets defined to.

`[^<]` matches anything not a `<`, including white spaces/newlines and therefore more thorough than `.`. The idea being that it would consume everything up to `location` and as long as that attribute was defined in the same `include` element.

I just find out attributes can be defined as child elements though, so I'll make an adjustment shortly.
I'll add comments to the code as well.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/LibertyWorkspace.java
cherylking: The dir scan on initialization sounds right, and adding the files within at that time. If they create a new file later on in one of these config dirs, will it be recognized as a Liberty config file? If not, is there a reasonable way for our extension to figure that out?

dshimo: I realized a simpler solution to be to simply check the file path lazily to see if it contains a config dir.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyConstants.java
```diff
@@ -22,9 +24,13 @@ private LibertyConstants() {
 
     public static final String FEATURE_MANAGER_ELEMENT = "featureManager";
     public static final String FEATURE_ELEMENT = "feature";
+    public static final String INCLUDE_ELEMENT = "include";
 
     public static final String PUBLIC_VISIBILITY = "PUBLIC";
 
     public static final String DEFAULT_SERVER_VERSION = "20.0.0.9";
 
+    public static final String WLP_USER_CONFIG_DIR = String.join(File.separator, "usr", "shared", "config");
```
cherylking, line 33: Does the resulting String start with or end with a file separator? The way you are using these constants now to determine if a file contains a config dir path, they should really have a file separator at both the beginning and end.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/LibertyWorkspace.java
cherylking: I like the new approach.

PR #59, opened 2022-05-03
Generate schema.xsd from ws-schemagen.jar

mattbsox:
Lemminx will now generate a schema file using `ws-schemagen.jar` in a Liberty installation if one exists for a workspace.

The generated file will be located in `workspaceDir/.libertyls`.

A cached schema file will be used if no Liberty installation is found.

Currently only one schema file will be generated per VSCode workspace.


lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyXSDURIResolver.java
```diff
@@ -69,4 +101,48 @@ public Map<String, String> getExternalGrammarLocation(URI fileURI) {
...
+          proc.destroy();
+          throw new Exception("Exceeded 30 second timeout during schema file generation. Using cached schema.xsd file.");
+        }
+  
+        LOGGER.info("Caching schema file with URI: " + xsdDestFile.toURI().toString());
+      } catch (Exception e) {
+        LOGGER.warning(e.getMessage());
+        return null;
+      }
+    }
```
cherylking, line 143: Might should log the fact that we are using a cached schema file here.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyXSDURIResolver.java
```diff
@@ -44,11 +48,39 @@ public class LibertyXSDURIResolver implements URIResolverExtension, IExternalGra
...
+          if (libertyWorkspace.isLibertyInstalled()) {
+            Path schemaGenJarPath = LibertyUtils.findFileInWorkspace(libertyWorkspace, "ws-schemagen.jar");
+            LOGGER.info("Generating schema file from: " + schemaGenJarPath.toString());
+            String serverSchemaUri = generateServerSchemaXsd(libertyWorkspace, schemaGenJarPath);
+            if (serverSchemaUri != null && !serverSchemaUri.isEmpty()) {
+              return serverSchemaUri;
+            }
+          }
+        }
+        Path serverXSDFile = CacheResourcesManager.getResourceCachePath(SERVER_XSD_RESOURCE);
```
cherylking, line 82: Should we log that we are using the default schema file and not a generated one here?

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyXSDURIResolver.java
```diff
@@ -69,4 +101,48 @@ public Map<String, String> getExternalGrammarLocation(URI fileURI) {
...
+   * @param libertyWorkspace
+   * @param schemaGenJarPath
+   * @return Path to generated schema file.
+   */
+  private String generateServerSchemaXsd(LibertyWorkspace libertyWorkspace, Path schemaGenJarPath) {
+    //java -jar path/to/ws-schemagen.jar path/to/workspace/.libertyls/libertySchema.xsd
+    File tempDir = LibertyUtils.getTempDir(libertyWorkspace.getWorkspaceURI().toString());
+
+    //TODO: (?) Add subfolders to tempDir: schema/xsd/liberty/server.xsd
+    File xsdDestFile = new File(tempDir, "server.xsd");
```
cherylking, line 115: Probably should guard against null here. The `getTempDir()` method has a couple of ways it can return null.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyXSDURIResolver.java
```diff
@@ -69,4 +101,48 @@ public Map<String, String> getExternalGrammarLocation(URI fileURI) {
...
+  
+        ProcessBuilder pb = new ProcessBuilder("java", "-jar", schemaGenJarPath.toAbsolutePath().toString(), xsdDestPath); //Add locale param here
+        pb.directory(tempDir);
+        pb.redirectErrorStream(true);
+        pb.redirectOutput(new File(tempDir, "schemagen.log"));
+  
+        Process proc = pb.start();
+        if (!proc.waitFor(30, TimeUnit.SECONDS)) {
+          proc.destroy();
+          throw new Exception("Exceeded 30 second timeout during schema file generation. Using cached schema.xsd file.");
```
cherylking, line 135: If this exception is thrown, the calling method does not default to using the cached schema.xsd. It probably should.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/LibertyProjectsManager.java
```diff
@@ -94,4 +94,7 @@ public void cleanUpTempDirs() {
         }
     }
 
+    public void cleanInstance() {
+        libertyWorkspaceFolders = new ArrayList<LibertyWorkspace>();
```
cherylking, line 98: Should this call `cleanUpTempDirs()` before reinitializing the array to empty?

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/SettingsService.java
```diff
@@ -56,6 +56,17 @@ public String getLibertyVersion() {
...
 
+  public String getLibertyRuntime() {
+    if (settings != null) {
+      String runtime = settings.getRuntime();
+      if (runtime != null) {
+        return runtime;
+      }
+    }
+
+    return null;
```
cherylking, line 67: This method seems inefficient. I think it can be reduced to the following.
```
return settings != null ? settings.getRuntime() : null;
```

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyUtils.java
```diff
@@ -104,6 +110,49 @@ public static Path findFileInWorkspace(String serverXmlURI, String filename) {
...
+        LibertyWorkspace libertyWorkspace = LibertyProjectsManager.getInstance().getWorkspaceFolder(serverXMLUri);
+
+        if (libertyWorkspace == null || libertyWorkspace.getWorkspaceURI() == null) {
+            return null;
+        }
+
+        String runtime = libertyWorkspace.getLibertyRuntime();
+
+        // return version from cache if set and Liberty is installed
+        if (runtime != null && libertyWorkspace.isLibertyInstalled()) {
```
cherylking, line 138: If the Liberty installation gets updated (they change the version in their pom.xml), will this get updated?

cherylking (changes requested):

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyXSDURIResolver.java
mattbsox: Added a new log message here too.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyXSDURIResolver.java
mattbsox: Removed the exception, added a log message, and return null here now.

The catch block was returning null when this exception was thrown, but it wasn't very clear.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/LibertyProjectsManager.java
mattbsox: Calling `cleanInstance()` in `stop()` in `LibertyExtension` after `cleanUpTempDirs()`.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/SettingsService.java
mattbsox: Fixed.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyUtils.java
mattbsox: It generates a new xsd file with the correct version, but the old version isn't deleted. Confirmed that it is using the new version.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyXSDURIResolver.java
mattbsox: Checking if tempDir is null, defaulting to the parent folder of ws-schemagen.jar. Added a null check for the ws-schemagen.jar as well.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyXSDURIResolver.java
```diff
@@ -111,37 +115,45 @@ private String generateServerSchemaXsd(LibertyWorkspace libertyWorkspace, Path s
     //java -jar path/to/ws-schemagen.jar path/to/workspace/.libertyls/libertySchema.xsd
     File tempDir = LibertyUtils.getTempDir(libertyWorkspace.getWorkspaceURI().toString());
 
+    //If tempDir is null, generate the schema file in the same directory as ws-schemagen.jar 
+    if (tempDir == null) {
+      tempDir = new File(schemaGenJarPath.toString()).getParentFile();
```
cherylking, line 120: Wouldn't this generate it in the Liberty installation itself? That doesn't seem like a good idea.

mattbsox: Yeah, it would. What would be a better alternative for a default? The `server.xml` location?

cherylking (approved):

PR #60, opened 2022-05-11
init Code Actions + Some Config File Additions

dshimo:
- [x] Code Actions

- Initialize new LibertyCodeActionParticipant in extension
- Construct map of error code to ICodeActionParticipant
- Provide CreateFile and EditAttribute as basic quick fixes
  - CreateFile will peek into the `<include>` node, grab the `location` attribute, and attempt to create a relative file from the current config file.

create file

  - EditAttribute will replace the text from the diagnostic range to `optional="true"`
  
Peek 2022-05-16 17-49

  - AddAttribute will insert `optional="true"` before `location`

add_attribute

- [x] Config File Changes

- Checks relative path, then absolute path, with server config dir check in part of a different call.
- URL content verification to be determined. So for this PR, diagnostic warning is muted based on heuristic if `location` attribute starts with `http`
- Similarly, env vars haven't be resolved, so muted with heuristic of containing `$`

- New Diagnostic message for when resource is not found and optional is set to false
image
and two quick fix solutions for either creating an empty server config file or changing optional value
image

- [x] OS compatibility

- Empty server config content gets initialized with `"<server>" + System.lineSeparator() + "</server>"`
- Relative path for new file is obtained from document URI as base, and resolved.

### DIFF
--- a/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
+++ b/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
@@ -137,7 +137,7 @@ public class LibertyDiagnosticParticipant implements IDiagnosticsParticipant {
             if (configFile.exists()) {
                 LibertyProjectsManager.getInstance().getWorkspaceFolder(domDocument.getDocumentURI()).addConfigFile(configFile.getCanonicalPath());
             } else {
-                if (optString != null && required) {
+                if (required) {
                     DOMNode optNode = node.getAttributeNode("optional");
                     Range optRange = XMLPositionUtility.createRange(optNode.getStart(), optNode.getEnd(), domDocument);
                     list.add(new Diagnostic(optRange, "The specified resource cannot be skipped. Check location value or set optional to true.", 
--- a/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
+++ b/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
@@ -28,17 +28,15 @@ import io.openliberty.tools.langserver.lemminx.util.*;
 
 import java.io.File;
 import java.io.IOException;
-import java.net.URI;
-import java.util.*;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
 
 public class LibertyDiagnosticParticipant implements IDiagnosticsParticipant {
 
     public static final String MISSING_FILE_CODE = "missing_file";
-    public static final String MISSING_FILE_MESSAGE = "The resource at the specified location could not be found:\n";
-
     public static final String NOT_OPTIONAL_CODE = "not_optional";
-    public static final String NOT_OPTIONAL_MESSAGE = "The specified resource cannot be skipped. Check location value or set optional to true.";
-
+    
     @Override
     public void doDiagnostics(DOMDocument domDocument, List<Diagnostic> diagnostics,
             XMLValidationSettings validationSettings, CancelChecker cancelChecker) {
@@ -59,7 +57,7 @@ public class LibertyDiagnosticParticipant implements IDiagnosticsParticipant {
             if (LibertyConstants.FEATURE_MANAGER_ELEMENT.equals(node.getNodeName())) {
                 validateFeature(domDocument, list, node);
             } else if (LibertyConstants.INCLUDE_ELEMENT.equals(node.getNodeName())) {
-                monitorConfigFiles(domDocument, list, node);
+                validateIncludeLocation(domDocument, list, node);
             }
         }
         
@@ -100,17 +98,29 @@ public class LibertyDiagnosticParticipant implements IDiagnosticsParticipant {
         }
     }
 
-    private void monitorConfigFiles(DOMDocument domDocument, List<Diagnostic> list, DOMNode node) {
+    /**
+     * Location checks:
+     * 1) Relative path; 2) Server config dir; 
+     * 3) Absolute path; 4) Web server
+     * 
+     * Checks in method: 1), 3)
+     * 2) performed in isConfigXMLFile
+     * 4) not yet implemented/determined
+     */
+    private void validateIncludeLocation(DOMDocument domDocument, List<Diagnostic> list, DOMNode node) {
         String locAttribute = node.getAttribute("location");
         if (locAttribute == null) {
+            return;
+        }
+        // skip diagnostic for not yet implemented behaviors/checks (URL + vars)
+        if (locAttribute.startsWith("http") || locAttribute.contains("$")) {
             return;
         }
 
         DOMNode locNode = node.getAttributeNode("location");
         Range range = XMLPositionUtility.createRange(locNode.getStart(), locNode.getEnd(), domDocument);
-
         if (!locAttribute.endsWith(".xml")) {
-            String message = "The specified file is not an XML file.";
+            String message = "The specified resource is not an XML file.";
             list.add(new Diagnostic(range, message, DiagnosticSeverity.Warning, "liberty-lemminx"));
             return;
         }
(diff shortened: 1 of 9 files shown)


lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
```diff
@@ -94,39 +98,57 @@ private void validateFeature(DOMDocument domDocument, List<Diagnostic> list, DOM
...
+        if (locAttribute.startsWith("http") || locAttribute.contains("$")) {
+            return;
+        }
 
         DOMNode locNode = node.getAttributeNode("location");
         Range range = XMLPositionUtility.createRange(locNode.getStart(), locNode.getEnd(), domDocument);
-
         if (!locAttribute.endsWith(".xml")) {
-            String message = "The specified file is not an XML file.";
+            String message = "The specified resource is not an XML file.";
```
cherylking, line 123: Just noting here that we will need to turn these into translatable messages once Kevin's merges his PR that has the Messages.java and some properties files.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
```diff
@@ -94,39 +98,57 @@ private void validateFeature(DOMDocument domDocument, List<Diagnostic> list, DOM
...
 
-        String docURIString = domDocument.getDocumentURI().replace(File.separator, "/");
-        locAttribute = locAttribute.replace(File.separator, "/");
-        File configFile = locAttribute.startsWith("./") ? 
-                new File(URI.create(docURIString.substring(0, docURIString.lastIndexOf("/") + 1))
-                        .resolve(locAttribute).normalize()) :
-                new File(locAttribute);
-
+        String optString = node.getAttribute("optional");
+        boolean required = optString == null ? false : optString.equals("false");
```
cherylking, line 129: Do we need to worry about trimming the attribute value that we get back before comparison?

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
```diff
@@ -94,39 +98,57 @@ private void validateFeature(DOMDocument domDocument, List<Diagnostic> list, DOM
...
+            configFile = new File(locAttribute);
+        }
         try {
             if (configFile.exists()) {
-                LibertyProjectsManager.getInstance().getWorkspaceFolder(docURIString).addConfigFile(configFile.getCanonicalPath());
+                LibertyProjectsManager.getInstance().getWorkspaceFolder(domDocument.getDocumentURI()).addConfigFile(configFile.getCanonicalPath());
             } else {
-                String message = "The file at the specified location could not be found.";
-                list.add(new Diagnostic(range, message, DiagnosticSeverity.Warning, "liberty-lemminx"));
+                if (optString != null && required) {
```
cherylking, line 140: Can't you just check `required` here? Whether the optional attribute was specified or not, if `required` is true this diagnostic needs to be flagged, right?

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/codeactions/CreateFile.java
```diff
@@ -0,0 +1,53 @@
...
+                new File(parentFile, locationValue).getCanonicalPath(), 
+                EMPTY_SERVER_CONFIG, diagnostic));
+
+            /* Uncomment to add option when the need is found */
+            // codeActions.add(CodeActionFactory.createFile(
+            //     "Create the missing server config file with location value as the absolute path.",
+            //     new File(locationValue).getCanonicalPath(), 
+            //     SERVER_TAGS_CONTENT, diagnostic));
+        } catch (Exception e) {
+            // BadLocationException not expected
```
cherylking, line 50: Are we eating an exception here if the file cannot be created?

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/codeactions/EditAttribute.java
```diff
@@ -0,0 +1,43 @@
...
+import org.eclipse.lsp4j.Diagnostic;
+import org.eclipse.lsp4j.Range;
+
+public class EditAttribute implements ICodeActionParticipant {
+
+    @Override
+    public void doCodeAction(Diagnostic diagnostic, Range range, DOMDocument document, List<CodeAction> codeActions,
+            SharedSettings sharedSettings, IComponentProvider componentProvider) {
+        try {
+            String title = "Change the optional attribute from false to true.";
```
cherylking, line 33: Is it optional by default, or required by default? I think the text could be simplified to "Set the optional attribute to true."

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/codeactions/EditAttribute.java
```diff
@@ -0,0 +1,43 @@
...
+            String replaceText = "optional=\"true\"";
+            codeActions.add(CodeActionFactory.replace(title, diagnostic.getRange(), replaceText, document.getTextDocument(), diagnostic));
+
+            // also build option to create file
+            new CreateFile().doCodeAction(diagnostic, range, document, codeActions, sharedSettings, componentProvider);
```
cherylking, line 38: why?

lemminx-liberty/src/test/java/io/openliberty/LibertyDiagnosticTest.java
```diff
@@ -80,32 +80,39 @@ public void testDiagnosticsForInclude() throws IOException {
         String serverXML = String.join(newLine, //
                 "<server description=\"default server\">", //
                 "    <include optional=\"true\" location=\"./empty_server.xml\"/>", //
-                "    <include optional=\"false\" location=\"MISSING FILE\"/>", //
+                "    <include optional=\"true\" location=\"/empty_server.xml\"/>", //
+                "    <include optional=\"true\" location=\"MISSING FILE\"/>", //
```
cherylking, line 84: Do you have any tests where the file is missing and the optional attribute is not specified?

cherylking (changes requested):

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/codeactions/EditAttribute.java
dshimo: I thought we discussed that if a file doesn't exist on path which is required, we could provide the user the options to either create the file or make the file not required?

lemminx-liberty/src/test/java/io/openliberty/LibertyDiagnosticTest.java
dshimo: Will add

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
cherylking: I was talking about the `optional` attribute value which is `true` or `false`. Not sure what `xml` has to do with it?

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/codeactions/CreateFile.java
cherylking: But we won't have any idea why the quick fix option is not added? Maybe we should at least log something?

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/codeactions/EditAttribute.java
cherylking: I guess I thought the code action to create the file would be on the location attribute, not the optional attribute. So it seemed odd that this code was here.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/codeactions/EditAttribute.java
cherylking: I was asking..which is it? If the `optional` attribute is not specified, does it default to `true` or `false`?

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
dshimo: We won't need to worry about values for `optional` as the schema defines it

cherylking: Can they specify `FALSE`? You said the schema defines the values, but does that mean we won't get anything besides `true` or `false` here?

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
```diff
@@ -94,39 +98,57 @@ private void validateFeature(DOMDocument domDocument, List<Diagnostic> list, DOM
...
+            configFile = new File(locAttribute);
+        }
         try {
             if (configFile.exists()) {
-                LibertyProjectsManager.getInstance().getWorkspaceFolder(docURIString).addConfigFile(configFile.getCanonicalPath());
+                LibertyProjectsManager.getInstance().getWorkspaceFolder(domDocument.getDocumentURI()).addConfigFile(configFile.getCanonicalPath());
             } else {
-                String message = "The file at the specified location could not be found.";
-                list.add(new Diagnostic(range, message, DiagnosticSeverity.Warning, "liberty-lemminx"));
+                if (required) {
```
cherylking, line 140: You said the optional flag defaults to `false` in another comment. So if it is null above at line 129, shouldn't `required` default to `true`?

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/codeactions/EditAttribute.java
cherylking: ok..I still think the text should change.

cherylking (changes requested):

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/codeactions/CreateFile.java
cherylking: Is logging possible here?

cherylking (approved):

Issue #50, opened 2022-03-24
Add support for bootstrap.properties and server.env


cherylking:
Recognize that the bootstrap.properties and server.env files are Liberty config files and apply the Liberty language server to them. 

File named “bootstrap.properties” and exists in either a “liberty/config” or “usr/servers” directory.
File named “server.env” and exists in either a “liberty/config” or “usr/servers” directory.

The first release of the language server will support common core properties and log and trace properties that can be specified in a server.env or bootstrap.properties file. Slides 29 - 32 of the [UFO](https://ibm.ent.box.com/file/802277171276?s=pfw3l83ny1gy3k11x1611uhfbxnyrta1) contain the list of properties/env variables to include.

1) Completion support
2) Hover support

cherylking:
For the hover support, let's put all hover text for these files in messages.properties and enable globalization.

PR #61, opened 2022-05-12
Initial hover support for bootstrap.properties and server.env

kevinwlau:
Initial hover support for #50

bootstrap.properties example: 
image

server.env example:
image


liberty-ls/src/main/java/io/openliberty/tools/langserver/resources/BootstrapPropertiesOptions.java
```diff
@@ -0,0 +1,62 @@
...
+
+public class BootstrapPropertiesOptions extends ListResourceBundle {
+ public static ResourceBundle serverenvProps = ResourceBundle.getBundle("io.openliberty.tools.langserver.resources.ServerEnvOptions");
+    
+ public Object[][] getContents() {
+ return resources;
+    }
+
+ private final static Object[][] resources = {
+        { "command.port", "Custom property. Overrides the default behavior of the Liberty server command port." },
```
cherylking, line 23: I  am pretty sure that the properties have to be in a `*.properties` file for translation. Those get shipped to translation centers and they return `*_<locale>.properties` files for each supported language.

cherylking (changes requested):

liberty-ls/src/main/java/io/openliberty/tools/langserver/model/propertiesfile/PropertiesEntryInstance.java
```diff
@@ -0,0 +1,63 @@
...
+        return position.getCharacter() <= propertyKeyInstance.getEndPosition();
+    }
+
+    public CompletableFuture<Hover> getHover(Position position) {
+        if (!isComment) {
+            if (isOnEntryKey(position)) {
+                return propertyKeyInstance.getHover();
+            } else {
+                return propertyValueInstance.getHover();
+            }
```
cherylking, line 59: Which does it return if the position is on the `=`?

liberty-ls/src/main/java/io/openliberty/tools/langserver/utils/Messages.java
```diff
@@ -0,0 +1,45 @@
...
+        if (serverenvMessages == null) {
+            initializeBundles();
+        }
+        String message = null;
+        if (serverEnvKeyFormat.matcher(key).matches()) { // server env
+            message = serverenvMessages.getString(key);
+        } else if (key.contains(".")) { // bootstrap property
+            message = bootstrapMessages.getString(key);
+            // 2nd lookup for properties that have a server.env equivalent
+            if (serverEnvKeyFormat.matcher(message).matches()) {
```
cherylking, line 39: Can you explain this logic a bit? If the description is not in bootstrapMessages, what does it return? How is that going to match something in serverEnvKeyFormat regex?

cherylking (changes requested):

liberty-ls/src/main/java/io/openliberty/tools/langserver/model/propertiesfile/PropertiesEntryInstance.java
kevinwlau: Hovering over the `=` returns the hover for the key. Wondering if this should be changed to just not show any hover?

liberty-ls/src/main/java/io/openliberty/tools/langserver/utils/Messages.java
kevinwlau: Currently, some of the entries in `BootstrapProperties.properties` are like this: `com.ibm.ws.logging.log.directory=LOG_DIR`, and `LOG_DIR` is defined in the ServerEnv.properties
If the value of the bootstrap matches the format for a server.env key, then it does a lookup on the ServerEnv properties to get its description.

For either server.env or bootstrap.properties, if the key is not found, then no hover appears. It currently throws a MissingResourceException for not being able to find the key, but doesn't break anything. May add in some exception handling for it.

liberty-ls/src/main/java/io/openliberty/tools/langserver/model/propertiesfile/PropertiesValueInstance.java
```diff
@@ -0,0 +1,40 @@
...
+        this.textDocumentItem = textDocumentItem;
+    }
+
+    public PropertiesKeyInstance getKey() {
+        return key;
+    }
+
+    public CompletableFuture<Hover> getHover() {
+        Hover hover = new Hover();
+        hover.setContents(new MarkupContent("plaintext", "This value is set for: " + this.key));
```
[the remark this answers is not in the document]
kevinwlau: It currently is, but now I'm reconsidering whether it's even necessary or helpful to show a hover for the value. There is no new information given.

liberty-ls/src/main/java/io/openliberty/tools/langserver/model/propertiesfile/PropertiesEntryInstance.java
cherylking: I think showing the hover for the key is fine. I just wanted to understand the behavior.

liberty-ls/src/main/java/io/openliberty/tools/langserver/model/propertiesfile/PropertiesValueInstance.java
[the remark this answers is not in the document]
cherylking: In the future we may resolve a value that has references to other variables/properties. For now, I'm not sure if the value should show any hover text. Maybe post the question in the slack channel.

liberty-ls/src/main/java/io/openliberty/tools/langserver/utils/Messages.java
cherylking: Ok, got it. Doesn't seem like throwing an exception when a key is not found makes sense though. There can definitely be keys that our LS will not recognize but would be perfectly valid.

liberty-ls/src/main/java/io/openliberty/tools/langserver/model/propertiesfile/PropertiesEntryInstance.java
```diff
@@ -0,0 +1,63 @@
...
+            isComment = true;
+            return;
+        }
+        isComment = false;
+        int equalsIndex = line.indexOf("=");
+        String propertyKeyInstanceString;
+        String propertyValueInstanceString;
+        if (equalsIndex != -1) {
+            propertyKeyInstanceString = line.substring(0, equalsIndex);
+            propertyValueInstanceString = line.substring(equalsIndex+1);
```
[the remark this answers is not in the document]
kevinwlau: Interesting catch, it would seem so. 
I looked at the javadocs for the `substring(beginIndex)` method, and an example seems to show that calling it on that index returns an empty string.
To test, I tried `+2` and that ended up out of bounds, but it's fine with the `+1`.

liberty-ls/src/main/java/io/openliberty/tools/langserver/utils/Messages.java
kevinwlau: The current Exception doesn't cause any problems with the language server continuing to run, it just provides some information on what key was not found.

I may add a try/catch for it to clean up the console output.

liberty-ls/src/main/java/io/openliberty/tools/langserver/model/propertiesfile/PropertiesEntryInstance.java
kevinwlau: I realized my method `getEndPosition` on the key actually returned the wrong value (off by 1). 
Fixed it to return the proper end index, so the hover no longer shows when hovering over `=`

liberty-ls/src/main/resources/BootstrapProperties.properties
```diff
@@ -0,0 +1,38 @@
...
+com.ibm.ws.logging.message.format=WLP_LOGGING_MESSAGE_FORMAT
+com.ibm.ws.logging.message.source=WLP_LOGGING_MESSAGE_SOURCE
+com.ibm.ws.logging.trace.file.name=This setting specifies the name of the trace log file. The trace.log file is created only if additional or detailed trace is enabled. `stdout` is recognized as a special value, and causes trace to be directed to the original standard out stream.
+com.ibm.ws.logging.trace.format=This setting controls the format of the trace log. The default format for Liberty is `ENHANCED`. You can also use `BASIC` and `ADVANCED` formats
+com.ibm.ws.logging.trace.specification=This setting is used to selectively enable trace. The log detail level specification is in the following format: `component = level`
+com.ibm.ws.logging.apps.write.json=WLP_LOGGING_APPS_WRITE_JSON
```
cherylking, line 31: I think this is going to cause issues with the translation folks. They won't understand how to translate this. We may just need to keep a Map in Messages.java that has this key/value pairing. If the key contains a `.` and is in that map, then get the value from the map and use it to retrieve the message from the other properties file.

cherylking (changes requested):

liberty-ls/src/main/resources/BootstrapProperties.properties
kevinwlau: I've gone ahead and implemented this locally, but I don't like how the bootstrap properties are split up between the `BootstrapProperties.properties` file and a separate Map that contains the equivalent property pairings.

cherylking (changes requested):

liberty-ls/src/main/resources/BootstrapProperties.properties
cherylking: Just recording here what we discussed in Slack. It will be less confusing for the translators if we move the special handling for these properties into Messages.java. That way we will not have messages in the properties file that should not be translated. These properties already had special behavior in Messages.java anyway.

cherylking (approved):

PR #63, opened 2022-05-16
Update workspaceURI check and reorganize schema gen logging

mattbsox:
Fixes issue with similar workspace folder names -> `test` and `test-20.0.0.12`

### DIFF
--- a/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/LibertyProjectsManager.java
+++ b/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/LibertyProjectsManager.java
@@ -12,6 +12,7 @@
 *******************************************************************************/
 package io.openliberty.tools.langserver.lemminx.services;
 
+import java.io.File;
 import java.io.IOException;
 import java.net.URI;
 import java.nio.file.Files;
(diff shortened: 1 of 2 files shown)


cherylking (approved):
Glad it was simple fix and that our discussion at scrum helped resolve this!

cherylking:
Seems there is a NPE in the windows build.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/LibertyProjectsManager.java
```diff
@@ -63,7 +64,13 @@ public String getLibertyVersion(LibertyWorkspace libertyWorkspace) {
      */
     public LibertyWorkspace getWorkspaceFolder(String serverXMLUri) {
         for (LibertyWorkspace folder : getInstance().getLibertyWorkspaceFolders()) {
-            if (serverXMLUri.contains(folder.getWorkspaceString())) {
+            //Append workspaceDirUri with file separator to avoid bad matches
+            String workspaceDirUri = folder.getWorkspaceString();
+            if (!workspaceDirUri.endsWith(File.separator)) {
```
cherylking, line 69: Can `workspaceDirUri` be null? Just wondering because of the NPE in the Windows build.

mattbsox: Yeah, some of the tests don't build the full workspace since the initializeParams are null.

mattbsox: Not sure if things would get this far since the workspace folders should be empty in that case.

mattbsox: Nevermind, looks like this test does set some workspaces, so it might be null there then.

PR #73, opened 2022-06-20
init config scan refactor

dshimo:


dshimo:
This was a mistake and unnecessary.

PR #79, opened 2022-07-25
Rearrange values for default. Preselect default value for completion

evie-lau:
Also added `TBASIC` as an option for message formats, according to [OpenLiberty docs](https://openliberty.io/docs/latest/log-trace-configuration.html) and [WAS Libertydocs](https://www.ibm.com/docs/en/was-liberty/base?topic=SSEQTP_liberty%2Fcom.ibm.websphere.liberty.autogen.nd.doc%2Fae%2Frwlp_config_logging.html)

### DIFF
--- a/liberty-ls/src/main/java/io/openliberty/tools/langserver/model/propertiesfile/PropertiesKeyInstance.java
+++ b/liberty-ls/src/main/java/io/openliberty/tools/langserver/model/propertiesfile/PropertiesKeyInstance.java
@@ -66,6 +66,11 @@ public class PropertiesKeyInstance {
     public List<CompletionItem> getValidValues() {
         List<String> values = ServerPropertyValues.getValidValues(textDocumentItem, propertyKey);
         List<CompletionItem> results = values.stream().map(s -> new CompletionItem(s)).collect(Collectors.toList());
+        // Preselect the default.
+        // This uses the first item in the List as default. 
+        // (Check ServerPropertyValues.java) Currently sorted to have confirmed/sensible values as default.
+        CompletionItem first = results.get(0);
+        first.setPreselect(true);
         return results;
     }
 
--- a/liberty-ls/src/main/java/io/openliberty/tools/langserver/utils/ServerPropertyValues.java
+++ b/liberty-ls/src/main/java/io/openliberty/tools/langserver/utils/ServerPropertyValues.java
@@ -26,24 +26,24 @@ public class ServerPropertyValues {
     private final static List<String> YES_NO_VALUES = Arrays.asList("y", "n");
     
     private static HashMap<String, List<String>> validServerValues = new HashMap<String, List<String>>() {{
-        put("WLP_DEBUG_SUSPEND", YES_NO_VALUES);
-        put("WLP_DEBUG_REMOTE", YES_NO_VALUES);
+        put("WLP_DEBUG_SUSPEND", YES_NO_VALUES); // default yes
+        put("WLP_DEBUG_REMOTE", YES_NO_VALUES); // default undefined/disabled, so it's pointless selecting no
         
-        put("WLP_LOGGING_CONSOLE_FORMAT", Arrays.asList("dev", "simple", "json"));
-        put("WLP_LOGGING_CONSOLE_LOGLEVEL", Arrays.asList("INFO", "AUDIT", "WARNING", "ERROR", "OFF"));
-        put("WLP_LOGGING_CONSOLE_SOURCE", LOGGING_SOURCE_VALUES);
-        put("WLP_LOGGING_MESSAGE_FORMAT", Arrays.asList("simple", "json"));
-        put("WLP_LOGGING_MESSAGE_SOURCE", LOGGING_SOURCE_VALUES);
-        put("WLP_LOGGING_APPS_WRITE_JSON", BOOLEAN_VALUES);
+        put("WLP_LOGGING_CONSOLE_FORMAT", Arrays.asList("DEV", "SIMPLE", "JSON", "TBASIC")); // default "DEV"
+        put("WLP_LOGGING_CONSOLE_LOGLEVEL", Arrays.asList("AUDIT", "INFO", "WARNING", "ERROR", "OFF")); // default "AUDIT"
+        put("WLP_LOGGING_CONSOLE_SOURCE", LOGGING_SOURCE_VALUES); //default "message"
+        put("WLP_LOGGING_MESSAGE_FORMAT", Arrays.asList("SIMPLE", "JSON", "TBASIC")); // default "SIMPLE"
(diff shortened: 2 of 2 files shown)


liberty-ls/src/main/java/io/openliberty/tools/langserver/utils/ServerPropertyValues.java
```diff
@@ -26,24 +26,24 @@ public class ServerPropertyValues {
...
-        put("com.ibm.ws.logging.newLogsOnStart", BOOLEAN_VALUES);
-        put("com.ibm.ws.logging.isoDateFormat", BOOLEAN_VALUES);
-        put("com.ibm.ws.logging.trace.format", Arrays.asList("ENHANCED", "BASIC", "ADVANCED"));
+        put("com.ibm.ws.logging.copy.system.streams", BOOLEAN_VALUES); // default true
+        put("com.ibm.ws.logging.newLogsOnStart", BOOLEAN_VALUES); // default true
+        put("com.ibm.ws.logging.isoDateFormat", Arrays.asList("false", "true")); // default false
+        put("com.ibm.ws.logging.trace.format", Arrays.asList("ENHANCED", "BASIC", "TBASIC", "ADVANCED")); // default "ENHANCED"
         put("websphere.log.provider", Arrays.asList("binaryLogging-1.0"));
-        put("com.ibm.hpel.log.bufferingEnabled", BOOLEAN_VALUES);
+        put("com.ibm.hpel.log.bufferingEnabled", BOOLEAN_VALUES); // default undefined/disabled
```
cherylking, line 46: This doesn't  make sense to me. A boolean property has to have a default. If this property is not specified, the server has a default behavior. What is it?

liberty-ls/src/main/java/io/openliberty/tools/langserver/utils/ServerPropertyValues.java
```diff
@@ -26,24 +26,24 @@ public class ServerPropertyValues {
...
-        put("WLP_LOGGING_CONSOLE_SOURCE", LOGGING_SOURCE_VALUES);
-        put("WLP_LOGGING_MESSAGE_FORMAT", Arrays.asList("simple", "json"));
-        put("WLP_LOGGING_MESSAGE_SOURCE", LOGGING_SOURCE_VALUES);
-        put("WLP_LOGGING_APPS_WRITE_JSON", BOOLEAN_VALUES);
+        put("WLP_LOGGING_CONSOLE_FORMAT", Arrays.asList("DEV", "SIMPLE", "JSON", "TBASIC")); // default "DEV"
+        put("WLP_LOGGING_CONSOLE_LOGLEVEL", Arrays.asList("AUDIT", "INFO", "WARNING", "ERROR", "OFF")); // default "AUDIT"
+        put("WLP_LOGGING_CONSOLE_SOURCE", LOGGING_SOURCE_VALUES); //default "message"
+        put("WLP_LOGGING_MESSAGE_FORMAT", Arrays.asList("SIMPLE", "JSON", "TBASIC")); // default "SIMPLE"
+        put("WLP_LOGGING_MESSAGE_SOURCE", LOGGING_SOURCE_VALUES); //default "message"
+        put("WLP_LOGGING_APPS_WRITE_JSON", BOOLEAN_VALUES); //unknown default
```
cherylking, line 37: This doesn't  make sense to me. A boolean property has to have a default. If this property is not specified, the server has a default behavior. What is it?

liberty-ls/src/main/java/io/openliberty/tools/langserver/model/propertiesfile/PropertiesKeyInstance.java
```diff
@@ -66,6 +66,11 @@ public CompletableFuture<List<CompletionItem>> getCompletions(Position position)
     public List<CompletionItem> getValidValues() {
         List<String> values = ServerPropertyValues.getValidValues(textDocumentItem, propertyKey);
         List<CompletionItem> results = values.stream().map(s -> new CompletionItem(s)).collect(Collectors.toList());
+        // Preselect the default.
+        // This uses the first item in the List as default. 
+        // (Check ServerPropertyValues.java) Currently sorted to have confirmed/sensible values as default.
+        CompletionItem first = results.get(0);
+        first.setPreselect(true);
```
cherylking, line 73: Does this mean the first item is selected even if you said there was no default (as in the case of WLP_DEBUG_REMOTE)?

cherylking (changes requested):

cherylking:
Found this page which seems to contain the default values for the `appsWriteJson` and `bufferingEnabled` properties. 

https://www.ibm.com/docs/en/was-liberty/base?topic=SSEQTP_liberty/com.ibm.websphere.liberty.autogen.nd.doc/ae/rwlp_config_logging.html

liberty-ls/src/main/java/io/openliberty/tools/langserver/model/propertiesfile/PropertiesKeyInstance.java
evie-lau: Yes, it will preselect the first item. This is where it gets a bit spicy. 

While technically there's no default for `WLP_DEBUG_REMOTE`, by not defining it, remote debugging is not allowed (essentially a `n` default). 
But reasonably, I think anyone adding this property would be using it to enable remote debugging, so they would want to define it with `y` which is the current preselected value.

Would it make more sense to preselect the default behavior? or the intended/desired behavior by someone adding this property?

liberty-ls/src/main/java/io/openliberty/tools/langserver/utils/ServerPropertyValues.java
```diff
@@ -23,6 +23,7 @@
 public class ServerPropertyValues {
     private final static List<String> LOGGING_SOURCE_VALUES = Arrays.asList("message", "trace", "accessLog", "ffdc", "audit");
     private final static List<String> BOOLEAN_VALUES = Arrays.asList("true", "false");
+    private final static List<String> BOOLEAN_VALUES_REVERSE = Arrays.asList("false", "true");
```
cherylking, line 26: Would make more sense to call these BOOLEAN_VALUES_DEFAULT_TRUE and BOOLEAN_VALUES_DEFAULT_FALSE.

liberty-ls/src/main/java/io/openliberty/tools/langserver/model/propertiesfile/PropertiesKeyInstance.java
cherylking: So you confirmed what I thought would happen, and it is a bit tricky which value to preselect. But I think it is reasonable to select `y`. Let's make it clear in the comment though that `y` will get preselected for this property.

cherylking (approved):

PR #87, opened 2022-08-22
Diagnostics for server.env and bootstrap.properties

evie-lau:
Add diagnostics for clearly defined property values in server.env and bootstrap.properties (defined in `ServerPropertyValues.java`)


liberty-ls/src/main/java/io/openliberty/tools/langserver/utils/PropertiesValidationResult.java
```diff
@@ -0,0 +1,65 @@
...
+        this.textDocumentItem = entry.getTextDocument();
+        this.hasErrors = false;
+    }
+
+    public void validateServerProperty() {
+        if (entry.isComment() || !ServerPropertyValues.canValidateKeyValues(textDocumentItem, entry.getKey())) {
+            return;
+        }
+        List<String> validValues = ServerPropertyValues.getValidValues(textDocumentItem, entry.getKey());
+        if (!validValues.stream().anyMatch(entry.getValue()::equalsIgnoreCase)) { // check if validValues contains the property value, ignoring case
```
cherylking, line 40: Does Liberty ignore the case for all the pre-defined values for the various properties/env variables?

liberty-ls/src/main/resources/DiagnosticMessages.properties
```diff
@@ -0,0 +1 @@
+INVALID_PROPERTY_VALUE=The value `{0}` is not valid for the property `{1}`.
```
cherylking, line 1: The bootstrap.properties contains properties, but the server.env contains variables. I think we need a slightly different diagnostic messages for the two files, unless we can come up with a generic term that makes sense.

For server.env: The value `{0}` is not valid for the variable `{1}`.

cherylking (changes requested):

liberty-ls/src/main/java/io/openliberty/tools/langserver/utils/PropertiesValidationResult.java
cherylking: We are following up on this. Apparently some props/vars are case sensitive, and others are not. Once we have identified all the recognized values, the code can be updated for this PR.

evie-lau: Updated the code to follow the current state of case-sensitivity for the properties/variables.
In example: 
image

liberty-ls/src/main/java/io/openliberty/tools/langserver/diagnostic/LibertyPropertiesDiagnosticService.java
```diff
@@ -0,0 +1,107 @@
...
+        Map<String, PropertiesValidationResult> errors = new HashMap<>();
+        if (openedDocument.getUri().endsWith("bootstrap.properties") || openedDocument.getUri().endsWith("server.env")) {
+            BufferedReader br = new BufferedReader(new StringReader(text));
+            String line = null;
+            int lineNumber = 0;
+            try {
+                while ((line=br.readLine()) != null) {
+                    PropertiesValidationResult validationResult = ServerPropertyValues.validateServerProperty(line, openedDocument);
+                    if (validationResult.hasErrors()) {
+                        validationResult.setLineNumber(lineNumber);
```
cherylking, line 52: Does the IDE expect line number zero for the first line of the file?

liberty-ls/src/main/java/io/openliberty/tools/langserver/utils/PropertiesValidationResult.java
cherylking: Should the test case be tweaked to cover case sensitive vs not?

liberty-ls/src/main/java/io/openliberty/tools/langserver/diagnostic/LibertyPropertiesDiagnosticService.java
evie-lau: Yes, it is zero-indexed, different from what the user sees in the IDE line numbers.

liberty-ls/src/main/java/io/openliberty/tools/langserver/utils/PropertiesValidationResult.java
evie-lau: Currently the language server portion of the project does not have any tests. I can look into setting up a testing framework for the project next.

cherylking: k...that is a whole we will want to fill. Go ahead and open an issue to follow up on adding tests for the LS.

cherylking (approved):

PR #93, opened 2022-08-26
Diagnostics part 2

### LINKED ISSUE #75: Add diagnostic support for server.env and bootstrap.properties
Add diagnostic/validation for properties in server.env and bootstrap.properties

For example:
- ~~invalid value (for those with well defined set of values)~~ - [Completed] #87
- invalid value for incorrect type (ie. integers, possibly within a certain range such as ports) such as `default.http.port` and `default.https.port`

evie-lau:
Resolves #75 
Add diagnostics for:
- [x] properties with integer values
- [x] org.osgi.framework.bootdelegation (comma separated list of packages)


liberty-ls/src/main/java/io/openliberty/tools/langserver/utils/PropertiesValidationResult.java
```diff
@@ -32,18 +38,79 @@ public PropertiesValidationResult(PropertiesEntryInstance entry) {
...
+                    hasErrors = !range.contains(Integer.parseInt(value));
+                } catch (NumberFormatException e) {
+                    hasErrors = true;
+                }
+            }
+            diagnosticType = ParserFileHelperUtil.isBootstrapPropertiesFile(textDocumentItem) ? 
+                                            "INVALID_PROPERTY_INTEGER_RANGE" : "INVALID_VARIABLE_INTEGER_RANGE";
+        } else if (ServerPropertyValues.usesPackageNames(property)) { // defines packages
+            // simple check for comma-delimited list of Java packages
+            Pattern packageList = Pattern.compile("^([a-z]+(\\.[a-z][a-z0-9]*)*)(,[a-z]+(\\.[a-z][a-z0-9]*)*)*$");
```
cherylking, line 108: According to this [page](https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html), the underscore `_` is also allowed.

liberty-ls/src/main/java/io/openliberty/tools/langserver/utils/ServerPropertyValues.java
```diff
@@ -72,20 +73,50 @@ public class ServerPropertyValues {
         );
     }};
 
-    public static boolean canValidateKeyValues(LibertyTextDocument tdi, String key) {
+    // TODO: it seems specifying Integers above MAX_VALUE breaks all diagnostics...
```
cherylking, line 76: Is this TODO still true? Or did you manage to add the various `catches` that were needed and handle the exceptions appropriately?

liberty-ls/src/main/java/io/openliberty/tools/langserver/diagnostic/LibertyPropertiesDiagnosticService.java
```diff
@@ -76,8 +76,9 @@ private List<Diagnostic> computeInvalidValuesDiagnostic(PropertiesValidationResu
             String lineContentInError) {
         List<Diagnostic> lspDiagnostics = new ArrayList<>();
         if (validationResult.hasErrors()) {
-            String messageTemplate = DiagnosticMessages.getString(validationResult.getInvalidValueMessageTemplate());
-            String message = MessageFormat.format(messageTemplate, validationResult.getValue(), validationResult.getKey());
+            String property = validationResult.getKey();
+            String messageTemplate = DiagnosticMessages.getString(validationResult.getDiagnosticType());
+            String message = MessageFormat.format(messageTemplate, validationResult.getValue(), property, ServerPropertyValues.getIntegerRange(property));
```
cherylking, line 81: What if this property does not have a property range? Does it just pass null and it is ignored because the message does not expect a value? Might be worth a comment in the code explaining.

cherylking (changes requested):

liberty-ls/src/main/java/io/openliberty/tools/langserver/utils/ServerPropertyValues.java
evie-lau: This TODO has been resolved 👍🏼

liberty-ls/src/main/java/io/openliberty/tools/langserver/diagnostic/LibertyPropertiesDiagnosticService.java
evie-lau: Yes, I was looking into reworking this, but currently the `{2}` arg is only used by the Integer messages. I'll add the comment explaining it

dshimo (approved):

cherylking (approved):

PR #31, opened 2020-11-04
Support document link for include elements

rzgry:
Adds support for document links for the location attribute of `<include>` elements. 

Fixes #19 

Screen Recording 2020-11-04 at 1 59 52 PM

PR #96, opened 2022-09-02
Document links for include elements

dshimo:
Pasting in work made by @rzgry in PR #31 

Also made minor corrections to devc support in `watchFiles` and test

### DIFF
--- a/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDocumentLinkParticipant.java
+++ b/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDocumentLinkParticipant.java
@@ -1,0 +1,59 @@
+package io.openliberty.tools.langserver.lemminx;
+
+import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.stream.Collectors;
+
+import org.apache.xerces.impl.XMLEntityManager;
+import org.apache.xerces.util.URI.MalformedURIException;
+import org.eclipse.lemminx.dom.DOMAttr;
+import org.eclipse.lemminx.dom.DOMDocument;
+import org.eclipse.lemminx.dom.DOMNode;
+import org.eclipse.lemminx.services.extensions.IDocumentLinkParticipant;
+import org.eclipse.lemminx.utils.XMLPositionUtility;
+import org.eclipse.lsp4j.DocumentLink;
+import org.eclipse.lsp4j.Range;
+
+import io.openliberty.tools.langserver.lemminx.util.LibertyUtils;
+
+
+public class LibertyDocumentLinkParticipant implements IDocumentLinkParticipant {
+
+    private static final Logger LOGGER = Logger.getLogger(LibertyDocumentLinkParticipant.class.getName());
+
+    @Override
+    public void findDocumentLinks(DOMDocument document, List<DocumentLink> links) {
+
+        if (!LibertyUtils.isConfigXMLFile(document))
+            return;
+
+        List<DOMNode> childNodes = document.getDocumentElement().getChildren();
+
+        // collect all <include> nodes that are children of the document element
+        List<DOMNode> includeDomNodes = childNodes.stream().filter(n -> n.getNodeName().equals("include"))
+                .collect(Collectors.toList());
+
+        for (DOMNode includeNode : includeDomNodes) {
+            DOMAttr includeAttr = includeNode.getAttributeNode("location");
+            Range linkRange = XMLPositionUtility.selectAttributeValue(includeAttr);
+            try {
+                String location = getResolvedLocation(document.getDocumentURI(), includeAttr.getValue());
+                DocumentLink link = new DocumentLink(linkRange, location);
+                links.add(link);
+            } catch (MalformedURIException e) {
+                LOGGER.log(Level.SEVERE, "Creation of document link failed", e);
+            }
+            
+        }
+
+    }
+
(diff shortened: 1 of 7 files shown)


lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDocumentLinkParticipant.java
```diff
@@ -0,0 +1,63 @@
+package io.openliberty.tools.langserver.lemminx;
```
cherylking, line 1: copyright

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDocumentLinkParticipant.java
```diff
@@ -0,0 +1,63 @@
...
+    @Override
+    public void findDocumentLinks(DOMDocument document, List<DocumentLink> links) {
+
+        if (!LibertyUtils.isConfigXMLFile(document))
+            return;
+
+        List<DOMNode> childNodes = document.getDocumentElement().getChildren();
+
+        // collect all <include> nodes that are children of the document element
+        List<DOMNode> includeDomNodes = childNodes.stream().filter(n -> n.getNodeName().equals("include"))
```
cherylking, line 34: I feel like there is a lot of similar code here that is also in https://github.com/OpenLiberty/liberty-language-server/blob/main/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java. Can these two classes share some logic? Maybe put it in LibertyUtils? Or make a new util these two classes utilize?

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDocumentLinkParticipant.java
```diff
@@ -0,0 +1,63 @@
...
+            DOMAttr includeAttr = includeNode.getAttributeNode("location");
+            
+            String locAttr = includeAttr.getValue();
+            if (!locAttr.endsWith(".xml") || locAttr.startsWith("http") || locAttr.contains("$")) {
+                // ignoring http or $ vars until feature supported
+                continue;
+            }
+            Range linkRange = XMLPositionUtility.selectAttributeValue(includeAttr);
+            try {
+                String location = getResolvedLocation(document.getDocumentURI(), includeAttr.getValue());
```
cherylking, line 47: So do we know it is a valid location at this point? Or do we go into the `catch` below if it is not?

lemminx-liberty/src/test/java/io/openliberty/LibertyWorkspaceTest.java
```diff
@@ -17,17 +17,17 @@
...
         LibertyWorkspace libertyWorkspace = new LibertyWorkspace(resourcesDir.toString());
         assertNull(libertyWorkspace.getContainerName());
         assertFalse(libertyWorkspace.isContainerAlive());
-        
-        assertNotNull(libertyWorkspace.findDevcMetadata());
-        
-        assertEquals("liberty-dev", libertyWorkspace.getContainerName());
-        assertTrue(libertyWorkspace.isContainerAlive());
-    }
+        assertNull(libertyWorkspace.findDevcMetadata());    // no alive containers return null
```
cherylking, line 26: I'm confused by the test case changes. The test metadata file has a containerName `liberty-dev`, so why is line 24 checking that it is null? Is that because you changed the metadata to say the container is not alive?

cherylking (changes requested):

lemminx-liberty/src/test/java/io/openliberty/LibertyWorkspaceTest.java
dshimo: Yes, and `findDevcMetadata` returns the first alive file or null

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDocumentLinkParticipant.java
dshimo: We go into the `catch` if the URI couldn't be expanded/generated for the link. We don't particularly test for validity -- diagnostics does that

cherylking: When you did the demo, it seemed like the document link only got created when it was an actual valid location. That is why I asked the question. I think your answer is saying it would go into the `catch` if it was not a valid location. Once the document link is created, I understand from your demo that the location could be changed to something invalid, and then we rely on our diagnostics to flag that. Did I get that right?

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDocumentLinkParticipant.java
cherylking: After looking at it again, there are some commonalities, but not enough to make refactoring worthwhile.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDocumentLinkParticipant.java
dshimo: Ah, yes

cherylking (approved):

PR #98, opened 2022-09-06
Trim feature name for hover and diagnostics

### LINKED ISSUE #64: Diagnostic message for feature with trailing space is identical to msg for non-existent feature
For a feature with trailing space like

```xml
    <featureManager>
        <feature>cdi-2.0 </feature>
```

I get a message that it doesn't exist:

> {"jsonrpc":"2.0","method":"textDocument/publishDiagnostics","params":{"uri":"file:///C:/ydosppa/asi2/src/main/liberty/config/server.xml","diagnostics":[{"range":{"start":{"line":18,"character":17},"end":{"line":18,"character":25}},"severity":1,"source":"liberty-lemminx","message":"ERROR: The feature \"cdi-2.0 \" does not exist."}]}}

This is confusing since the runtime tolerates this by stripping the space, leaving me not realizing the extra space is the issue.   

Should the message be customized for this case?  Can it be downgraded in severity too (haven't really looked at the diagnostic part of the LSP in detail but I see the severity property in the payload).

mattbsox:
Fixes #64

### DIFF
--- a/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
+++ b/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
@@ -81,7 +81,7 @@ public class LibertyDiagnosticParticipant implements IDiagnosticsParticipant {
             DOMNode featureTextNode = (DOMNode) featureNode.getChildNodes().item(0);
             // skip nodes that do not have any text value (ie. comments)
             if (featureTextNode != null && featureTextNode.getTextContent() != null) {
-                String featureName = featureTextNode.getTextContent();
+                String featureName = featureTextNode.getTextContent().trim();
                 // if the feature is not a user defined feature and the feature does not exist in the list of
                 // supported features show a "Feature does not exist" diagnostic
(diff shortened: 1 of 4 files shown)


lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
```diff
@@ -81,7 +81,7 @@ private void validateFeature(DOMDocument domDocument, List<Diagnostic> list, DOM
             DOMNode featureTextNode = (DOMNode) featureNode.getChildNodes().item(0);
             // skip nodes that do not have any text value (ie. comments)
             if (featureTextNode != null && featureTextNode.getTextContent() != null) {
-                String featureName = featureTextNode.getTextContent();
+                String featureName = featureTextNode.getTextContent().trim();
```
cherylking, line 84: Is there a test case that can be updated to show this works?

mattbsox: Yeah, I can modify one.

cherylking (approved):

PR #105, opened 2022-09-15
Remove licenses and html files and merge notices

### LINKED ISSUE #102: Legal info (notices, licenses, etc.) gets lost in 'jar-with-dependencies' build of liberty-langserver-lemminx extension
Because of the approach used to build the **liberty-langserver-lemminx-1.0-SNAPSHOT-jar-with-dependencies.jar**, using the assembly plugin with     `<descriptorRef>jar-with-dependencies</descriptorRef>` config, the contents of the various dependency JARs get smashed together and overwritten.

This makes it harder to consume this JAR downstream, e.g. in the Liberty Tools Eclipse feature, where we have to identify the legal info and bubble it up to the users of our feature.   

E.g. if I build the lemminx-liberty module I see these license/notice files:
```
> 
> META-INF/LICENSE.txt
> META-INF/LICENSE.md
> META-INF/NOTICE.md

```

It is unclear which dependency each of these came from, since they have been smashed together.

I think we should either do a better job at preserving the individual parts or just remove them altogether.   

The latter is the approach we took in the Liberty Tools Eclipse Early Release .  If you look at 
https://github.com/OpenLiberty/liberty-tools-eclipse/blob/main/bundles/io.openliberty.tools.eclipse.lsp4e/lib/liberty-langserver-lemminx-1.0-SNAPSHOT-jar-with-dependencies.jar you'll see we removed these and rely on https://github.com/OpenLiberty/liberty-tools-eclipse/blob/main/bundles/io.openliberty.tools.eclipse.lsp4e/about.html to track the info.

That could work.  Or if you want to look at preserving them, renaming them or adding to a different per-dependency dir, that could work too.

cherylking:
Fixes #102 

The license transformer removes the license files (could not find a way to move to a subfolder per dependency). I am also removing html files so that we don't have multiple about*.html files in the `jar-with-dependencies` and merging the NOTICES.

### DIFF
--- a/lemminx-liberty/pom.xml
+++ b/lemminx-liberty/pom.xml
@@ -45,20 +45,41 @@
                 </configuration>
             </plugin>
             <plugin>
-                <artifactId>maven-assembly-plugin</artifactId>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-shade-plugin</artifactId>
                 <version>3.3.0</version>
-                <configuration>
-                    <descriptorRefs>
-                        <descriptorRef>jar-with-dependencies</descriptorRef>
-                    </descriptorRefs>
-                </configuration>
                 <executions>
                     <execution>
-                        <id>make-assembly</id> <!-- this is used for inheritance merges -->
+                        <id>make-assembly</id>
                         <phase>package</phase> <!-- bind to the packaging phase -->
                         <goals>
-                            <goal>single</goal>
+                            <goal>shade</goal>
                         </goals>
+                        <configuration>
+                            <finalName>${project.artifactId}-${project.version}-jar-with-dependencies</finalName>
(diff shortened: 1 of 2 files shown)


cherylking:
Here are the contents for the lemminx jar with dependencies META-INF folder:

Before: 
Screen Shot 2022-09-15 at 3 01 51 PM

After: 
Screen Shot 2022-09-15 at 3 01 42 PM

cherylking:
Here are the contents for the liberty-ls jar with dependencies:

Before: 
Screen Shot 2022-09-15 at 3 08 24 PM

After:
Screen Shot 2022-09-15 at 3 09 05 PM

cherylking:
@scottkurz Can you approve this PR? Are the changes sufficient?

cherylking:
If more changes are needed, a separate issue can be open. Will go ahead and merge this.

Issue #89, opened 2022-08-24
Enhance tests beyond basic functionality


evie-lau:
The language server features (hover, completion, diagnostics) for server.env and bootstrap.properties files need tests.

cherylking:
Going to have David look into this since @evie-lau is still working on the diagnostics PR.

evie-lau:
As of PR #108 and #112, all features (hover, completion, diagnostics) are tested.
Consider adding some tests for the Language Server itself, or for the TextDocumentService, where these tests try all the features at once. ie: check that completion works and has the property description, or hovering over the diagnostic gives the correct message.

cherylking:
Closing this as the original request in the issue has been completed.

PR #108, opened 2022-09-27
Tests and completion updates

### LINKED ISSUE #107: Build should include liberty-ls
The current Github actions are only building lemminx-liberty and not liberty-ls. We should build both when PRs are created/modified.

evie-lau:
Some work for #89 and resolves #107 

- Modify Github Actions to run the build for liberty-ls
- Add tests for hover and completion features for liberty-ls
- Modify completion behavior a bit to filter results on language server-side, instead of relying completely on client-side automatic filtering.

### DIFF
--- a/liberty-ls/src/main/java/io/openliberty/tools/langserver/LibertyLanguageServer.java
+++ b/liberty-ls/src/main/java/io/openliberty/tools/langserver/LibertyLanguageServer.java
@@ -23,13 +23,15 @@ import org.eclipse.lsp4j.InitializedParams;
 import org.eclipse.lsp4j.ServerCapabilities;
 import org.eclipse.lsp4j.TextDocumentSyncKind;
 import org.eclipse.lsp4j.services.LanguageClient;
+import org.eclipse.lsp4j.services.LanguageClientAware;
 import org.eclipse.lsp4j.services.LanguageServer;
 import org.eclipse.lsp4j.services.TextDocumentService;
 import org.eclipse.lsp4j.services.WorkspaceService;
 
 
-public class LibertyLanguageServer implements LanguageServer {
+public class LibertyLanguageServer extends AbstractLanguageServer implements LanguageServer, LanguageClientAware {
 
+    public static final String LANGUAGE_ID = "LANGUAGE_ID_LIBERTY";
     private static final Logger LOGGER = Logger.getLogger(LibertyLanguageServer.class.getName());
 
     private final WorkspaceService workspaceService;
@@ -77,13 +79,19 @@ public class LibertyLanguageServer implements LanguageServer {
     }
 
     @Override
+    public void connect(LanguageClient client) {
+        this.languageClient = client;
+    }
+
+    @Override
     public CompletableFuture<Object> shutdown() {
-        // TODO Auto-generated method stub
-        return null;
+        super.shutdownServer();
+        return CompletableFuture.completedFuture(new Object());
     }
 
     @Override
     public void exit() {
+        super.stopServer();
         System.exit(0);
     }
 
--- a/liberty-ls/src/main/java/io/openliberty/tools/langserver/utils/Messages.java
+++ b/liberty-ls/src/main/java/io/openliberty/tools/langserver/utils/Messages.java
@@ -82,8 +82,14 @@ public class Messages {
         }
         
         String filename = textDocument.getUri();
-        // remove completion results that don't contain the query string
-        Predicate<String> filter = s -> (!s.contains(query));
+        // remove completion results that don't contain the query string (case-insensitive search)
+        Predicate<String> filter = s -> {
+            for (int i = s.length() - query.length(); i >= 0; --i) {
+                if (s.regionMatches(true, i, query, 0, query.length()))
+                    return false;
+            }
+            return true;
+        };
         if (filename.contains("server.env")) { // server.env file
             List<String> keys = new ArrayList<String>(serverPropertyKeys);
             keys.removeIf(filter);
--- a/liberty-ls/src/main/java/io/openliberty/tools/langserver/utils/Messages.java
+++ b/liberty-ls/src/main/java/io/openliberty/tools/langserver/utils/Messages.java
@@ -9,11 +9,13 @@
 *******************************************************************************/
 package io.openliberty.tools.langserver.utils;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
 import java.util.MissingResourceException;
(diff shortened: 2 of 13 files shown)


liberty-ls/src/main/java/io/openliberty/tools/langserver/LibertyLanguageServer.java
```diff
@@ -23,13 +23,15 @@
...
 import org.eclipse.lsp4j.TextDocumentSyncKind;
 import org.eclipse.lsp4j.services.LanguageClient;
+import org.eclipse.lsp4j.services.LanguageClientAware;
 import org.eclipse.lsp4j.services.LanguageServer;
 import org.eclipse.lsp4j.services.TextDocumentService;
 import org.eclipse.lsp4j.services.WorkspaceService;
 
 
-public class LibertyLanguageServer implements LanguageServer {
+public class LibertyLanguageServer extends AbstractLanguageServer implements LanguageServer, LanguageClientAware {
```
cherylking, line 32: Will these changes affect how the various IDEs start/interact with the Liberty Language Server?

liberty-ls/src/main/java/io/openliberty/tools/langserver/LibertyLanguageServer.java
```diff
@@ -76,14 +78,20 @@ public synchronized void updateSettings(Object initializationOptionsSettings) {
         // TODO: else update settings
     }
 
+    @Override
```
cherylking, line 81: Are these really overrides? I don't see method connect, shutdown or exit in the class that is being extended. I see these were already overrides before, but I think that is incorrect. Also, is there a need to use `super` when the method is not in the current class and is inherited?

liberty-ls/src/main/java/io/openliberty/tools/langserver/utils/Messages.java
```diff
@@ -80,10 +82,22 @@ public static List<String> getMatchingKeys(String query, LibertyTextDocument tex
         }
         
         String filename = textDocument.getUri();
+        // remove completion results that don't contain the query string (case-insensitive search)
```
cherylking, line 85: I don't really follow what this is for. I read the comment, but that is not telling me what the query string is for (example of something we are removing). What is the use case?

cherylking (changes requested):

liberty-ls/src/main/java/io/openliberty/tools/langserver/utils/Messages.java
evie-lau: Some background... the initial implementation never filtered any results for completion, just returned the entire catalog of keys, and let the client-side IDE handle the filtering based on what was typed. 
When it came to running tests, each completion request would return that full list, making the tests pretty useless other than checking that: yes, we return all the possible keys for this file.

evie-lau: Functionally when running the IDE extension with the language server, it doesn't seem to make much of a difference. But adding this logic to filter the results on the server-side made it a lot easier to test reasonable functionality for the completion feature.

liberty-ls/src/main/java/io/openliberty/tools/langserver/LibertyLanguageServer.java
evie-lau: I don't think they will change. They should still use the original initialization methods we had in place before, and these additions are in place for our test suite to startup the language server.

liberty-ls/src/main/java/io/openliberty/tools/langserver/LibertyLanguageServer.java
evie-lau: Hmm, you are right, they're not overriding any superclass methods, but these methods are implementing the interface methods.
I think it is still a good marker to show it is an implemented method, even if it's not necessary. (I also read that Eclipse auto-marks implemented methods with `@Override` with Java 1.6+)
`super` is also not necessary, and serves more as an indicator. Should I remove these?

cherylking: k...if Eclipse auto-marks them that is fine. Just leave it.

liberty-ls/src/main/java/io/openliberty/tools/langserver/utils/Messages.java
cherylking: So it filters out based on a key? Or it is filtering out bootstrap vs server.env? Still not clear.

liberty-ls/src/main/java/io/openliberty/tools/langserver/LibertyLanguageServer.java
cherylking: The functionality looks different in the original initialization methods. That is why I asked this question.

liberty-ls/src/main/java/io/openliberty/tools/langserver/LibertyLanguageServer.java
```diff
@@ -76,14 +78,20 @@ public synchronized void updateSettings(Object initializationOptionsSettings) {
...
+    public void connect(LanguageClient client) {
+        this.languageClient = client;
+    }
+
     @Override
     public CompletableFuture<Object> shutdown() {
-        // TODO Auto-generated method stub
-        return null;
+        super.shutdownServer();
+        return CompletableFuture.completedFuture(new Object());
```
cherylking, line 89: The contents of the shutdown and exit method look pretty different to me from before. This is why I thought behavior might be different?

liberty-ls/src/main/java/io/openliberty/tools/langserver/utils/Messages.java
evie-lau: Yes, it filters the available keys based on the entered text (query).

liberty-ls/src/main/java/io/openliberty/tools/langserver/LibertyLanguageServer.java
evie-lau: You may be looking at the initialization for the Test language server?

liberty-ls/src/main/java/io/openliberty/tools/langserver/LibertyLanguageServer.java
evie-lau: Ah I see. The updated methods are just to ensure the language server instances from the tests are closed out.

cherylking (approved):

Issue #100, opened 2022-09-09
Change the schema generation to put it in the target/build directory


cherylking:
Currently, the generated schema file and associated log end up in the project directory under a `.libertyls` directory. Should generate the schema file in the appropriate `target` / `build` directory instead. We are already creating the devc metadata file in the `target` / `build` directory.

There are a couple of reasons to do this:

1) The file won't get tracked as a source code change.
2) When a `clean` is done the file will get regenerated.

We should move the devc metadata file under `.libertyls` directory in the target/build dir along with the schema file so they are all grouped together.

cherylking:
Here is a screenshot of the location of the files in a VS Code project.

![image (15)](https://user-images.githubusercontent.com/29436451/189400735-3e121a1e-d33d-4618-a19e-018805927f4d.png)

PR #109, opened 2022-09-29
Generate .libertyls folder in workspace buildDir

mattbsox:
Addresses #100

### DIFF
--- a/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyUtils.java
+++ b/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyUtils.java
@@ -219,7 +219,7 @@ public class LibertyUtils {
                 
                 if (devcOn) {
                     DockerService docker = DockerService.getInstance();
-                    File containerPropertiesFile = new File(getTempDir(libertyWorkspace.getWorkspaceString()), "container.properties");
+                    File containerPropertiesFile = new File(getTempDir(libertyWorkspace), "container.properties");
                     docker.dockerCp(libertyWorkspace.getContainerName(), DockerService.DEFAULT_CONTAINER_OL_PROPERTIES_PATH.toString(), containerPropertiesFile.toString());
                     fis = new FileInputStream(containerPropertiesFile);
                 } else {
@@ -253,21 +253,21 @@ public class LibertyUtils {
      * Return temp directory to store generated feature lists and schema. Creates
      * temp directory if it does not exist.
      * 
-     * @param folder WorkspaceFolderURI indicates where to create the temporary
-     *               directory
+     * @param libertyWorkspace LibertyWorkspace where the temporary directory is made
+     *               
      * @return temporary directory File object
      */
-    public static File getTempDir(String workspaceFolderURI) {
-        if (workspaceFolderURI == null) {
+    public static File getTempDir(LibertyWorkspace libertyWorkspace) {
+        if (libertyWorkspace == null) {
             return null;
         }
         try {
-            URI rootURI = new URI(workspaceFolderURI);
-            Path rootPath = Paths.get(rootURI);
-            File file = rootPath.toFile();
-            File libertyLSFolder = new File(file, ".libertyls");
-
-            if (!libertyLSFolder.exists()) {
+            File libertyLSFolder = new File(libertyWorkspace.getDir(), ".libertyls"); //Default to workspaceDir/.libertyls
+            Path pluginConfigFilePath = findFileInWorkspace(libertyWorkspace,Paths.get("liberty-plugin-config.xml"));
+            if (pluginConfigFilePath != null) { //If liberty-plugin-config.xml exists use its parent directory: buildDir/.libertyls
+                libertyLSFolder = new File(pluginConfigFilePath.getParent().toFile(), ".libertyls");
+            }
+            if (libertyLSFolder != null && !libertyLSFolder.exists()) {
                 if (!libertyLSFolder.mkdir()) {
                     return null;
                 }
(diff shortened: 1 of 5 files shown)


lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyUtils.java
```diff
@@ -253,19 +253,24 @@ public static String getVersion(String serverXMLUri) {
...
+        if (libertyWorkspace == null) {
             return null;
         }
         try {
-            URI rootURI = new URI(workspaceFolderURI);
-            Path rootPath = Paths.get(rootURI);
-            File file = rootPath.toFile();
-            File libertyLSFolder = new File(file, ".libertyls");
+            Path targetFolderPath = findFileInWorkspace(libertyWorkspace, Paths.get("target"));
+            Path buildFolderPath  = findFileInWorkspace(libertyWorkspace, Paths.get("build"));
```
cherylking, line 266: This seems too generic and error prone to me. It would match any folder ending with `target` or `build` wouldn't it? So if a workspace had a folder name `dev-target` it would match? Or `testbuild`?

cherylking: How do we find the installed Liberty to use the schema gen jar?

cherylking (changes requested):

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyUtils.java
cherylking: What if we looked for the `liberty-plugin-config.xml` file and used the parent dir of that? I have a feeling we will want to use values from that file down the road anyhow.

mattbsox: Yeah, I think that makes more sense. I'll update it to look for that file.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyUtils.java
```diff
@@ -253,19 +253,24 @@ public static String getVersion(String serverXMLUri) {
...
+
+            File libertyLSFolder = null;
+            if (targetFolderPath!= null && targetFolderPath.toFile().exists()) {
+                libertyLSFolder = new File(targetFolderPath.toFile(), ".libertyls");
+            } else if (buildFolderPath!= null && buildFolderPath.toFile().exists()) {
+                libertyLSFolder = new File(buildFolderPath.toFile(), ".libertyls");
+            }
 
             if (!libertyLSFolder.exists()) {
```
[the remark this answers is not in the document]
mattbsox: Yeah, but I think we would return null in that case anyway. I'll add a check. Shouldn't change the outcome there.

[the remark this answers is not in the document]
cherylking: Returning null would cause the calling methods to fail. Previously it could not return null. Perhaps we create in the old location if target/build are not found?

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyUtils.java
```diff
@@ -253,21 +253,18 @@ public static String getVersion(String serverXMLUri) {
...
+    public static File getTempDir(LibertyWorkspace libertyWorkspace) {
+        if (libertyWorkspace == null) {
             return null;
         }
         try {
-            URI rootURI = new URI(workspaceFolderURI);
-            Path rootPath = Paths.get(rootURI);
-            File file = rootPath.toFile();
-            File libertyLSFolder = new File(file, ".libertyls");
+            File libertyLSFolder = new File(findFileInWorkspace(libertyWorkspace,Paths.get("liberty-plugin-config.xml")).getParent().toFile(), ".libertyls");
```
cherylking, line 265: If the file is not found, will this get a NPE on the call to `getParent()`?

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyUtils.java
```diff
@@ -253,21 +253,18 @@ public static String getVersion(String serverXMLUri) {
...
 
-            if (!libertyLSFolder.exists()) {
+            if (libertyLSFolder != null && !libertyLSFolder.exists()) {
```
cherylking, line 267: I think we should go back to the old default location if the plugin file is not found.

cherylking (changes requested):

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyUtils.java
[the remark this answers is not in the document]
cherylking: Actually, it did return null before but only when the workspace was null.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyUtils.java
mattbsox: Should I keep this null check?

cherylking: Yes, but if it is null use the old default location to create the `.libertyls` folder.

cherylking (approved):

PR #112, opened 2022-10-04
Diagnostic tests

evie-lau:
More tests for #89. Adds tests for the diagnostic feature


cherylking (approved):
Looks good. I see the current tests are checking the correct number of diagnostic messages are triggered, and occur on the correct lines/index. Are you going to add a couple that check the actual messages that are sent?

evie-lau:
I'll add some of those as well

PR #114, opened 2022-10-07
Install shaded artifact

cherylking:
The latest snapshot builds only published the original `jar` and not the `jar-with-dependencies`. The Jenkins job publishes the installed artifacts from the build. When I switched to using the `maven-shade-plugin`, it caused the `jar-with-dependencies` to not get installed. Adding some configuration to the `maven-shade-plugin` to cause it to get installed.

Note: I tried several combinations to get the shaded jar as the only installed artifact. They all had undesirable side affects (the installed artifact name was not what I wanted no matter what config I tried). Even though we currently do not know of any need for the original `jar`, it will cause less churn to leave it as is.

### DIFF
--- a/lemminx-liberty/pom.xml
+++ b/lemminx-liberty/pom.xml
@@ -56,7 +56,8 @@
                             <goal>shade</goal>
                         </goals>
                         <configuration>
-                            <finalName>${project.artifactId}-${project.version}-jar-with-dependencies</finalName>
+                            <shadedArtifactAttached>true</shadedArtifactAttached>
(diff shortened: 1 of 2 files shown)


mattbsox (approved):
Looks good.

One more thing to note is that the `jar-with-dependencies` classifier needs to be added to the coordinates when specifying it as a dependnecy:
```
    <dependency>
        <groupId>io.openliberty.tools</groupId>
        <artifactId>liberty-langserver</artifactId>
        <version>1.0-SNAPSHOT</version>
        <classifier>jar-with-dependencies</classifier>
    </dependency>
```

PR #128, opened 2022-10-18
Fix feature list cache

### LINKED ISSUE #125: Fix feature list generation
This code does not seem correct. Any generated file should be in the `.libertyls` folder under the build/target folder. Second, I'm not sure what purpose the empty featurelist.xml serves. Either there should be no default feature list, or use a generated one.

https://github.com/OpenLiberty/liberty-language-server/blob/bb3a4dc34250a900fbd55cef434fa5277a707f0d/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/FeatureService.java#L213

cherylking:
Fixes #125 

1) Updated key for features cache to include runtime (`ol` or `wlp`).
2) Updated retrieval of features.json from Maven Central to use correct coordinates for `wlp` vs `ol`.
3) Updated feature list generation logic (which will only ever run if the features.json cannot be downloaded from Maven Central) to generate the feature list in the `.libertyls` folder in the build/target.
4) If the features.json cannot be downloaded, and if the feature list cannot be generated, then the default features.json is used. This language server currently includes the `22.0.0.9` features.json.

Trace showing that the features.json is downloaded and cached correctly:

Screen Shot 2022-10-17 at 5 00 51 PM

Screenshot of generated feature list xml file (versions `18.0.0.1` and older do not have a `features.json` available in Maven Central.

Screen Shot 2022-10-17 at 9 54 13 PM


lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/FeatureService.java
```diff
@@ -76,10 +80,10 @@ private FeatureService() {
    * @param libertyVersion - version of Liberty to fetch features for
    * @return list of features supported by the provided version of Liberty
    */
-  private List<Feature> fetchFeaturesForVersion(String libertyVersion) throws IOException, JsonParseException {
-    String featureEndpoint = String.format(
-        "https://repo1.maven.org/maven2/io/openliberty/features/features/%s/features-%s.json", libertyVersion,
-        libertyVersion);
+  private List<Feature> fetchFeaturesForVersion(String libertyVersion, String libertyRuntime) throws IOException, JsonParseException {
+    String featureEndpoint = libertyRuntime.equals("wlp") ? String.format(wlpFeatureEndpoint, libertyVersion, libertyVersion) :
```
evie-lau, line 84: I think the *FeatureEndpoint strings can be modified so we only need to pass `libertyVersion` once instead of twice in the `String.format()` args

kathrynkodama (commented):
LGTM

evie-lau (changes requested):

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/FeatureService.java
cherylking: How? That part of the code did not change. So I was not looking at how to optimize it.

evie-lau: `%1$s` should do it

evie-lau (approved):

Issue #135, opened 2022-11-15
README ID review


dmuelle:

dmuelle:
Review completed and edits implemented in #136. Closing as completed

PR #136, opened 2022-11-15
Initial ID review suggestions and questions

dmuelle:
#135


README.md
```diff
@@ -1,59 +1,78 @@
...
 * [Liberty Tools for Eclipse](https://github.com/OpenLiberty/liberty-tools-eclipse)
 * [Liberty Tools for Intellij](https://github.com/OpenLiberty/liberty-tools-intellij)
 
+## Supported files
+
+Liberty Config Language Server features are avaialble for the following configuration files.
+
+- `server.env`
+- `bootstrap.properties`
+- `server.xml` (or any XML file) and any XML files that are referenced through the `include` element in the `server.xml` file.
```
dmuelle, line 19: this bullet is a little confusing. I'm not sure what to make of "or any XML file", esp since we set limitations both in this sentence and in the sub bullet. What do we want to let users know by saying "any XML file" ?

Also

> XML files must contain the `server` root element and exist in the...

Does this also apply to included files?

README.md
```diff
@@ -1,59 +1,78 @@
...
+If the [Liberty Maven Plugin](https://github.com/OpenLiberty/ci.maven) or [Liberty Gradle Plugin](https://github.com/OpenLiberty/ci.gradle) is configured with the Liberty project, Liberty Config Language Server automatically generates a schema file based on the Liberty runtime and version to provide relevant information about the supported `server.xml` elements and Liberty features. A minimum version of the Liberty Maven Plugin 3.7.1 or Liberty Gradle Plugin 3.5.1 is recommended.  If an earlier version of either plugin is used, the schema file will not be regenerated when additional features are installed. This limitation might result in lack of hover, completion, or diagnostic support for elements related to those newly installed features.
```
dmuelle, line 69: it's not clear how this section fits in with the previous information- is this discussing a separate set of features from those already mentioned? Or do all features depend on this schema generation?

I can't tell if this is just to identify recommended minimum levels for all features, or just some subset that pertains to dev mode and maven

dmuelle:
Initial ID review, made some suggestions and have a few questions.

README.md
cherylking: Basically, we recognize `server.xml` files, but also xml files by other names when they meet the criteria that followed (located in a certain directory and have a `server` root element).

README.md
cherylking: All the support for `server.xml` (and other xml files that are Liberty server configuration as defined earlier) depend on the schema generation. If the schema cannot be generated, then a default one is used (which currently matches 22.0.0.9 - wonder if we should mention that?).

README.md
```diff
@@ -1,59 +1,78 @@
...
 * [Liberty Tools for Intellij](https://github.com/OpenLiberty/liberty-tools-intellij)
 
+## Supported files
+
+Liberty Config Language Server features are avaialble for the following configuration files.
+
+- `server.env`
+- `bootstrap.properties`
+- `server.xml` (or any XML file) and any XML files that are referenced through the `include` element in the `server.xml` file.
+    - XML files must contain the `server` root element and exist in the `src/main/liberty/config`, `configDropins/overrides`, `configDropins/defaults`, `usr/shared/config`, or `usr/servers` directory.
```
cherylking, line 20: Instead of making this a sub-bullet for the previous one, perhaps it should just be its own bullet:
- XML files that contain a `server` root element and are located in the ..., ..., ..., or ... directory.

Then change the previous bullet to:
- `server.xml` file and files referenced by an `include` element in the `server.xml` file.

README.md
```diff
@@ -1,59 +1,78 @@
...
+![Screen capture of a documentation dialog appearing when hovering over a Liberty variable in a server.env file](./docs/images/variable-hover.png "Hover on Liberty server variables in server.env")
+
 * Hover for Liberty XML configs
 
-![Screenshot of feature documentation appearing when hovering over a Liberty feature in a server.xml file](./docs/images/feature-hover.png "Hover on Liberty features in server.xml")
+![Screen capture of feature documentation appearing when hovering over a Liberty feature in a server.xml file](./docs/images/feature-hover.png "Hover on Liberty features in server.xml")
 
 ### Diagnostics on Liberty server configuration files
+
+Diagnostics highlight potential problems in your configuration, such as invlaid values.
```
cherylking, line 57: fix spelling `invlaid` to `invalid`

cherylking (changes requested):
Thanks for the review. Changes look good. Just had a couple of suggestions and one typo. Let me know if I answered your questions sufficiently.

README.md
dmuelle: do files that are referenced through the `include` element in the `server.xml` file also need to meet the root element and file location criteria?

README.md
dmuelle: ok- I'm going to revise my suggestion

README.md
dmuelle: good idea- I'll make that change

README.md
cherylking: They need the root `server` element, but they can be located anywhere.

cherylking (approved):

PR #143, opened 2022-12-06
Create NOTICES file and have build delete intermediate NOTICE and leg…

### LINKED ISSUE #131: Remove NOTICE/LICENSE from Liberty Lemminx Extension JAR
While at first, on paper the shade plugin transformers:
```xml
<transformer implementation="org.apache.maven.plugins.shade.resource.ApacheLicenseResourceTransformer"/>
<transformer implementation="org.apache.maven.plugins.shade.resource.ApacheNoticeResourceTransformer">
```
seem to align with what we'd want, I don't think it's working at the moment, and I think we might even be better off starting fresh like we do in Liberty Tools Eclipse.  

Mark this as a placeholder to follow-up.

scottkurz:
…al files and zip up this one

Signed-off-by: Scott Kurz <skurz@us.ibm.com>

Fixes #131

This PR takes the Liberty Tools Eclipse legal information we'd been maintaing here: 
https://github.com/OpenLiberty/liberty-tools-eclipse/blob/main/bundles/io.openliberty.tools.eclipse.lsp4e/about.html

and tries to refactor it to fit into the format of the Open Liberty runtime:
https://github.com/OpenLiberty/open-liberty/blob/integration/dev/build.image/publish/NOTICES
(which is a generated file).

I'm not saying this is a very common structure, but it's maybe one way to do it.  

I do agree that the Apache license & notice resource transformers initially seem to solve a problem that sounds like the one we're solving.   I'm not sure how to take them further though.   I'm not sure they work on non-Apache licenses and notices either.   It seemed better to me just to delete the individual license/notice files that were dragged along from the prereqs and start top-down by looking at the POM dependency tree.

### DIFF
--- a/lemminx-liberty/pom.xml
+++ b/lemminx-liberty/pom.xml
@@ -34,6 +34,14 @@
     </distributionManagement>
 
     <build>
+        <resources>
+            <resource>
+                <directory>${project.basedir}/..</directory>
+                <includes>
+                    <include>NOTICES</include>
+                </includes>
+            </resource>
+        </resources>
         <plugins>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
@@ -72,12 +80,15 @@
                                 </filter>
                             </filters>
                             <transformers>
-                                <transformer implementation="org.apache.maven.plugins.shade.resource.ApacheLicenseResourceTransformer"/>
-                                <transformer implementation="org.apache.maven.plugins.shade.resource.ApacheNoticeResourceTransformer">
-                                    <addHeader>false</addHeader>
-                                </transformer>
                                 <transformer implementation="org.apache.maven.plugins.shade.resource.DontIncludeResourceTransformer">
-                                    <resource>.html</resource>
+                                    <resources>
+                                        <resource>.html</resource>
+                                        <resource>META-INF/NOTICE</resource>
+                                        <resource>META-INF/NOTICE.txt</resource>
+                                        <resource>META-INF/NOTICE.md</resource>
+                                        <resource>META-INF/LICENSE.txt</resource>
+                                        <resource>META-INF/LICENSE.md</resource>
+                                     </resources>
                                 </transformer>
                             </transformers>
                         </configuration>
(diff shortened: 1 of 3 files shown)


liberty-ls/pom.xml
```diff
@@ -107,12 +113,13 @@
...
                                 </transformer>
-                                <transformer implementation="org.apache.maven.plugins.shade.resource.ApacheLicenseResourceTransformer"/>
-                                <transformer implementation="org.apache.maven.plugins.shade.resource.ApacheNoticeResourceTransformer">
-                                    <addHeader>false</addHeader>
-                                </transformer>
                                 <transformer implementation="org.apache.maven.plugins.shade.resource.DontIncludeResourceTransformer">
-                                    <resource>.html</resource>
+                                    <resources>
+                                        <resource>.html</resource>
+                                        <resource>META-INF/NOTICE</resource>
```
cherylking, line 119: Just wondering why you didn't include as many formats for the notice and license files here as you did for lemminx-liberty?

cherylking (commented):
Thanks for the PR Scott. Just had one question.

liberty-ls/pom.xml
cherylking: I think excluding the same list of files for each makes more sense.

lemminx-liberty/pom.xml
```diff
@@ -34,6 +34,14 @@
     </distributionManagement>
 
     <build>
+        <resources>
```
cherylking, line 37: I think I figured out the problem. When you added this resource section, it caused it to stop including the `src/main/resources` which happens by default. Need to add this section here:

```
            <resource>
                <directory>src/main/resources</directory>
            </resource>
```

cherylking (changes requested):

lemminx-liberty/pom.xml
mattbsox: This fix worked for me locally as well.

liberty-ls/pom.xml
scottkurz: ok..done

cherylking (approved):

PR #144, opened 2022-12-06
Minor README edits

yeekangc:
Updated IDE names

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,8 +7,8 @@ The Liberty Config Language Server provides language server features for Liberty
 The Liberty Config Language Server is included in the following client IDEs.
 
 * [Liberty Tools for VS Code](https://github.com/OpenLiberty/liberty-tools-vscode)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,8 +7,8 @@ The Liberty Config Language Server provides language server features for Liberty
 The Liberty Config Language Server is included in the following client IDEs.
 
 * [Liberty Tools for VS Code](https://github.com/OpenLiberty/liberty-tools-vscode)
```
cherylking, line 9: I think this one should be `Liberty Tools for Visual Studio Code` also, right? Thanks for the PR. We didn't catch this earlier.

cherylking (changes requested):

README.md
cherylking: I will make this update after merging.

cherylking (approved):

PR #146, opened 2022-12-06
Add logic to handle multi-module for Maven projects

### LINKED ISSUE #139: Do not propose configuration
in the server.xml, added `jwtSso-1.0` feature, have following issues. 

Ctrl+space does not propose `jwtSso` tag
Screen Shot 2022-11-17 at 8 58 37 AM

placed cursor after jwtSso tag at line 53, no more good proposals, expect more/all proposals
Screen Shot 2022-11-17 at 9 02 06 AM

But, after save the server.xml and wait for server restart, the issues do not happen. How users know that? Have to start dev mode?

cherylking:
Fixes #139

### DIFF
--- a/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/LibertyProjectsManager.java
+++ b/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/LibertyProjectsManager.java
@@ -93,13 +93,10 @@ public class LibertyProjectsManager {
                         }
                         List<Path> serverXmlFile = LibertyUtils.findFilesInDirectory(nextChildDir, serverXmlPath);
                         if (!serverXmlFile.isEmpty()) {
-                            Path pomFile = Paths.get(nextChildDir.toUri().getPath(), "pom.xml");
-                            if (pomFile.toFile().exists()) {
-                                LibertyWorkspace libertyWorkspace = new LibertyWorkspace(lastChildDirPath);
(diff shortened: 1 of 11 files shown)


lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/LibertyProjectsManager.java
```diff
@@ -45,8 +47,63 @@ private LibertyProjectsManager() {
...
+                                        .filter(Files::isDirectory)
+                                        .collect(Collectors.toList());
+
+                    for (Path nextChildDir : childrenDirs) {
+                        lastChildDirPath = nextChildDir.toUri().toString().replace("///", "/");;
+                        if (nextChildDir.equals(workspacePath)) {
+                            continue;
+                        }
+                        List<Path> serverXmlFile = LibertyUtils.findFilesInDirectory(nextChildDir, serverXmlPath);
+                        if (!serverXmlFile.isEmpty()) {
```
scottkurz, line 95: Maybe we don't need the Maven check.  It sounds like this logic doesn't really intersect with the reactor or aggregate aspects of the multi-module build.  If like we were saying last week, the project scope is mainly about drawing a boundary around server config includes... that would seem to stop at the submodule level.. and it also doesn't seem to require anything in LMP not in LGP.

scottkurz (approved):
LGTM

mattbsox (approved):

PR #149, opened 2022-12-12
Change build command to include integration tests

cherylking:


.github/workflows/lemminx-liberty.yml
```diff
@@ -22,7 +22,7 @@ jobs:
           java-version: 17
       - name: Build Lemminx Liberty
         working-directory: ./lemminx-liberty
-        run: mvn -B package --file pom.xml
+        run: mvn clean install
       - name: Build Liberty Language Server
```
evie-lau, line 26: I noticed the first build skipped the LCLS build when the preceding lemminx build failed.

Can we try adding this?
```suggestion
      - name: Build Liberty Language Server
        if: success() || failure()
```

It will allow the step to run even if the previous Lemminx build fails

cherylking: Will try that later when I can confirm the behavior.

Issue #152, opened 2022-12-14
On Windows, server XML config files not detected even in specifically targeted directories


scottkurz:
## RECREATE

1. Install Liberty Tools Eclipse https://public.dhe.ibm.com/ibmdl/export/pub/software/openliberty/liberty-tools-eclipse/0.7.0/repository/
2. Create a file src/main/liberty/config/configDropins/defaults/my.xml
3. Open my.xml in the Generic Text Editor
4. Try to do Ctrl+Space completion and notice it doesn't provide Liberty server XSD aware answers, and/or you get a CodeLens-style prompt 

## EXPLANATION

The LibertyUtils method
```java
    public static boolean isConfigDirFile(String filePath) {
        return filePath.contains(LibertyConstants.WLP_USER_CONFIG_DIR) ||
                filePath.contains(LibertyConstants.SERVER_CONFIG_DROPINS_DEFAULTS) ||
                filePath.contains(LibertyConstants.SERVER_CONFIG_DROPINS_OVERRIDES);
    }
```
uses constants like:
`public static final String SERVER_CONFIG_DROPINS_DEFAULTS = File.separator + String.join(File.separator, "configDropins", "defaults") + File.separator;`

However, the file path is coming via other lsp layers from a file type URI/URL.  This  uses forward-slash  ('/') separators even on Windows.   I don't have a definitive reference but looking in the debugger this seems to clearly be the issue.   

So the backslash in the comparison string is causing the 'contains' test to return 'false'.

cherylking:
@dshimo This can be closed right? The fix went in on PR 153.

PR #153, opened 2023-01-11
simple fix for win/uri conflicts

dshimo:
simple fix for issue #152

### DIFF
--- a/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyConstants.java
+++ b/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyConstants.java
@@ -29,6 +29,6 @@ public final class LibertyConstants {
     public static final String PUBLIC_VISIBILITY = "PUBLIC";
 
     public static final String WLP_USER_CONFIG_DIR = "/usr/shared/config/";
-    public static final String SERVER_CONFIG_DROPINS_DEFAULTS = "/configDropins/default/";
+    public static final String SERVER_CONFIG_DROPINS_DEFAULTS = "/configDropins/defaults/";
     public static final String SERVER_CONFIG_DROPINS_OVERRIDES = "/configDropins/overrides";
 }
--- a/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyConstants.java
+++ b/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyConstants.java
@@ -28,7 +28,7 @@ public final class LibertyConstants {
 
     public static final String PUBLIC_VISIBILITY = "PUBLIC";
 
-    public static final String WLP_USER_CONFIG_DIR = File.separator + String.join(File.separator, "usr", "shared", "config") + File.separator;
-    public static final String SERVER_CONFIG_DROPINS_DEFAULTS = File.separator + String.join(File.separator, "configDropins", "defaults") + File.separator;
-    public static final String SERVER_CONFIG_DROPINS_OVERRIDES = File.separator + String.join(File.separator, "configDropins", "overrides") + File.separator;
+    public static final String WLP_USER_CONFIG_DIR = "/usr/shared/config/";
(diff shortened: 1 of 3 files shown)


lemminx-liberty/src/test/java/io/openliberty/LibertyWorkspaceTest.java
```diff
@@ -30,4 +30,22 @@ public void testReadDevcMetadata() throws URISyntaxException {
...
+
+        File mockXML = new File("src/test/resources/configDropins/defaults/my.xml");
+        String filePathString = mockXML.getCanonicalPath();
+        URI filePathURI = mockXML.toURI();
+
+        assertFalse(LibertyUtils.isConfigDirFile(filePathString));
+        assertTrue(LibertyUtils.isConfigDirFile(filePathURI.toString()));
+        // mock replacement
+        filePathString = filePathString.replace("\\", "/");
+        assertTrue(LibertyUtils.isConfigDirFile(filePathString));
```
cherylking, line 49: This confuses me. Why are you replacing backslash with forward slash before calling `isConfigDirFile`? Seems like we should be testing that if the input has backslashes, it can still deal with it correctly.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyConstants.java
```diff
@@ -28,7 +28,7 @@ private LibertyConstants() {
 
     public static final String PUBLIC_VISIBILITY = "PUBLIC";
 
-    public static final String WLP_USER_CONFIG_DIR = File.separator + String.join(File.separator, "usr", "shared", "config") + File.separator;
-    public static final String SERVER_CONFIG_DROPINS_DEFAULTS = File.separator + String.join(File.separator, "configDropins", "defaults") + File.separator;
-    public static final String SERVER_CONFIG_DROPINS_OVERRIDES = File.separator + String.join(File.separator, "configDropins", "overrides") + File.separator;
+    public static final String WLP_USER_CONFIG_DIR = "/usr/shared/config/";
```
cherylking, line 31: Add a comment here explaining why it is ok, and in fact required, that the forward slash is used here.

cherylking (changes requested):
General comment...update the copyright to add 2023 for any changed files.

lemminx-liberty/src/test/java/io/openliberty/LibertyWorkspaceTest.java
dshimo: This test is a bit of a hack I admit. I didn't want to setup a mock Liberty workspace and project manager to check this simple behavior. 
This test seeks to prove two things when a file is read on Windows:
- the URI is handled correctly without edit when the constants are set to strings with forward slashes
- `getCanonicalPath` will fail `isConfigDirFile` when read on Windows until handled with the replace

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyConstants.java
dshimo: The original intent of `File.separator` was to cover the slash differences between operating systems. With #152, it's revealed that backslash appears specifically on Windows paths and not URIs. If the backslash is a minor case and gets addressed with a `.replace`, we can make forward slash the standard.
This change makes the constants more readable and more direct, over concat and string joins.

lemminx-liberty/src/test/java/io/openliberty/LibertyWorkspaceTest.java
```diff
@@ -1,3 +1,16 @@
+/*******************************************************************************
```
cherylking, line 1: copyrights are not needed for tests

cherylking (changes requested):

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyConstants.java
cherylking: I meant add a comment to the code. And it can be shorter than what you put here.

cherylking (approved):

PR #154, opened 2023-01-11
Add whitespace check around equals sign for server.env file

### LINKED ISSUE #141: Diagnostic doesn't work properly with whitespace around the `=` sign
Diagnostic does not work with whitespace before `=`
image

Diagnostic includes whitespaces in the value when there is whitespace after `=`
image

Diagnostic works correctly when there is no whitespace around the `=` sign
image

**Note**: Might have to confirm with someone if whitespace is even allowed in the first place.

Have seen this issue present in both VSCode and IntelliJ

evie-lau:
Fixes #141


liberty-ls/src/main/resources/DiagnosticMessages.properties
```diff
@@ -2,4 +2,5 @@ INVALID_PROPERTY_VALUE=The value `{0}` is not valid for the property `{1}`.
 INVALID_VARIABLE_VALUE=The value `{0}` is not valid for the variable `{1}`.
 INVALID_PROPERTY_INTEGER_RANGE=The value `{0}` is not within the valid range `{2}` for the property `{1}`.
 INVALID_VARIABLE_INTEGER_RANGE=The value `{0}` is not within the valid range `{2}` for the variable `{1}`.
-INVALID_PACKAGE_LIST_FORMAT=This value must be a comma-delimited list of Java packages.
\ No newline at end of file
+INVALID_PACKAGE_LIST_FORMAT=This value must be a comma-delimited list of Java packages.
+SERVER_ENV_INVALID_WHITESPACE=There should be no whitespace surrounding the equals sign (=).
```
cherylking, line 6: I see you copied this from the IBM doc, but you added an 's' to 'equal'.  Probably should just say 'equal'.

cherylking (changes requested):
Other than the one minor comment, also should update the copyright year to 2023. If there is only one year, then add ", 2023" to it. If there are two years, change the second year to 2023.

cherylking:
As we discussed on scrum, there should be changes for handling blank spaces correctly for `bootstrap.properties` as well. So don't have this PR close issue #141.

evie-lau:
TODO: fix/redo how we compute the range for the diagnostic. The original design did not support marking the keys in diagnostics.
image

cherylking (approved):

evie-lau:
Now properly marks the whitespaces around the equal sign in server.env, and more robust than before.
image

liberty-ls/src/main/java/io/openliberty/tools/langserver/diagnostic/LibertyPropertiesDiagnosticService.java
```diff
@@ -82,30 +82,36 @@ private List<Diagnostic> computeInvalidValuesDiagnostic(PropertiesValidationResu
...
+        if (startChar != null && endChar != null) {
+            return new Range(new Position(lineNumber, startChar), new Position(lineNumber, endChar));
+        }
+
+        // otherwise calculate range
+        if (startChar != null) {
+            // currently unused. use either whole line or equal index
+            endChar = lineContentInError.length();
+        } else if (endChar != null) {
+            startChar = equalIndex + 1;
```
cherylking, line 111: I feel like lines 101 through 112 could be simplified to something like this:

```
Position start = startChar == null ? new Position(lineNumber, equalIndex +1) : new Position(lineNumber, startChar);
Position end = endChar == null ? new Position(lineNumber, lineContentInError.length()) : new Position(lineNumber, endChar);
return new Range(start, end);
```

I wasn't sure what it was supposed to do if both `startChar` and `endChar` were null? Is that possible? Your code did not handle that exactly. It would fall through and use the nulls I think.

liberty-ls/src/main/java/io/openliberty/tools/langserver/utils/PropertiesValidationResult.java
```diff
@@ -24,7 +24,7 @@
 import io.openliberty.tools.langserver.model.propertiesfile.PropertiesEntryInstance;
 
 public class PropertiesValidationResult {
-    Integer lineNumber;
+    Integer lineNumber, startChar = null, endChar = null;
```
cherylking, line 27: Should `lineNumber` be initialized to null?

cherylking (changes requested):

liberty-ls/src/main/java/io/openliberty/tools/langserver/utils/PropertiesValidationResult.java
evie-lau: Another method had set `lineNumber` if the `hasErrors` flag was true. 
I changed the validation entry point to always assign a line number instead.

cherylking (approved):

PR #162, opened 2023-02-28
Changed logging fine to info

### LINKED ISSUE #138: Lemminx extension logging can't be enabled with current Lemminx Eclipse (and IDEA?) integration
Though we have **java.util.logging** instrumentation in the lemminx-liberty extension, there doesn't seem to be a way to enable it via Liberty Tools Eclipse integration.  

Some of the reason is discussed here: https://github.com/eclipse/wildwebdeveloper/discussions/974.

I was able to prototype a couple workarounds in the near term: 

### IDEA 1
1. Programmatically add ConsoleHandler to the root Logger, e.g. in a static/init block:   `Logger.getLogger("").addHandler(new ConsoleHandler());`
2. If we decide this should only be done conditionally, we can key off of a sysprop with a name like:  `io.openliberty.tools.lcls.logging.proxy=true`.   Why "proxy"?   Well, as it turns out, the Eclipse WWD integration will pass through properties with "proxy" in the name from the IDE config (e.g. eclipse.ini) to the XML LSP JVM.  So that would "work".
3. "Promote" everything we want to see to INFO level.  (Or do this programmatically but only conditionally along with the enablement itself in 2.)   

Note this could be noisy, I haven't tried it.  On the other hand, I'm not sure we've got the perfect balance as-is...e.g. the log message `"Using cached Liberty schema file located at: " + serverXSDFile.toString())`  gets logged an awful lot.  Not sure how much time's been spent getting the message levels right. 

### IDEA 2
1. Tell users to run with env var LEMMINX_DEBUG=true
2. Replace JUL calls with `System.err.println(msg)` (we can't just use System.out since Lemminx sets that to a NoOp stream).
3. As with IDEA 1, if this is conditional we need to use a sysprop with 'proxy' in the name

The 2nd ends up a good bit noisier since we're opening this up to all of Lemminx, e.g. the m2e in Eclipse gives quite a few stack traces.

cherylking:
Fixes #138

### DIFF
--- a/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyExtension.java
+++ b/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyExtension.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright (c) 2020, 2022 IBM Corporation and others.
+* Copyright (c) 2020, 2023 IBM Corporation and others.
 *
 * This program and the accompanying materials are made available under the
 * terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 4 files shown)


scottkurz (approved):
LGTM

PR #169, opened 2023-04-18
Changes for 0.25.0 version of Eclipse LemMinX

### LINKED ISSUE #163: Comments throws off diagnostic (cvc-complex-type.2.3) mark onto the wrong location in server.xml
The expected behavior is for the diagnostics/markers to appear on the text. This behavior is also observed in `pom.xml`
image

When a comment is in between the text and the `server` element, the diagnostic marks the comment on line 4 rather than line 12.
image
This issue does not seem to occur on `pom.xml`

Steps to recreate:
1. Create a `server.xml` with a `server` element
2. Put a comment with content into it
3. Type any text anywhere in the `server` element

This bug appears in IntelliJ and VSCode 
image

cherylking:
Fixes #163

### DIFF
--- a/lemminx-liberty/pom.xml
+++ b/lemminx-liberty/pom.xml
@@ -10,7 +10,7 @@
     <url>https://github.com/OpenLiberty/liberty-language-server</url>
 
     <properties>
-        <lemminx.version>0.22.0</lemminx.version>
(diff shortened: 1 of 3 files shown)


dshimo (approved):

cherylking:
Do NOT merge. This has breaking changes. Need to determine release strategy.

Issue #184, opened 2023-06-15
.libertyls and the related schema xsd fail to generate


dshimo:
Between now and the December early release, LCLS seems to fail to extract the schema out of a running Liberty container. LCLS functionality is questionable with some seeing support and some not.

The process works past generating the container metadata and has a hiccup after. 
LibertyXSDURIResolver.java would be a good place to look.

scottkurz:
I noticed this on Win 11.  Thinking it may be just a Windows issue...   the stack here:
> DockerService.generateServerSchemaXsdFromContainer(LibertyWorkspace) line: 100	
> LibertyXSDURIResolver.resolve(String, String, String) line: 84	

is building a command from a Windows Path:

```java
public static final Path DEFAULT_CONTAINER_SCHEMAGEN_JAR_PATH =
            DEFAULT_CONTAINER_WLP_DIR.resolve(Paths.get("bin", "tools", "ws-schemagen.jar"));
```

But it's a Linux container even though I"m running Windows.. so the Path is wrong.  We end up throwing an exception and returning `null` from **io.openliberty.tools.langserver.lemminx.LibertyXSDURIResolver.resolve(String, String, String)**

cherylking:
@dshimo Were you on Windows when you recreated it?

dshimo:
I could reproduce this on Mac.

cherylking:
> I could reproduce this on Mac.

Ok, so the exception Scott saw is a different root cause than yours on Mac. Can you find out what exception is happening on Mac?

cherylking:
@dshimo Any update on this? What exception is happening when running on mac?

dshimo:
First, there seems to be multiple factors involved as there were varying results among the IDE's and OS's. And to clarify, my initial reproduction was on IntelliJ + Mac.

|   | VS Code | IntelliJ | Eclipse |
| --- | --- | --- | --- |
| Windows | `.libertyls` gets generated, but Docker returns with exit code 1. Results in using cached features | Not yet tested | Scott's result
| Mac | Works as intended | `.libertyls` does not get generated (1) | Struggling to get working, both with plugin dev and marketplace install (2)

(1) When trying to debug IntelliJ, the code doesn't seem to go through LibertyXSDUriResolver which contains our logic. It travels through LSPTextHover for hover information. This could be an entirely different behavior from Scott's Windows path issue.

[...]

PR #187, opened 2023-07-17
devc pathing

dshimo:
Addresses the OS pathing portion of #184. 
DockerService uses `Path.resolve` to generate paths to use in a container. This fails on Windows as a Windows path gets built for devc, which are based on Linux.

- Known paths for Docker and Liberty are hardcoded
- Minor additions were added to log messages.

Verified in VS Code on Windows, before and after path changes.
image

### DIFF
--- a/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyXSDURIResolver.java
+++ b/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyXSDURIResolver.java
@@ -91,7 +91,7 @@ public class LibertyXSDURIResolver implements URIResolverExtension, IExternalGra
                 }
             }
         } catch (Exception e) {
-            LOGGER.severe("Error: Unable to deploy server.xsd to LemMinX cache.");
+            LOGGER.severe("Error: Unable to generate the Liberty schema from the target Liberty runtime.");
             e.printStackTrace();
         }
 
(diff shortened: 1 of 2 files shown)


cherylking (commented):
I think we should change the error handling in the class/method that actually calls `generateServerSchemaXsdFromContainer` so that we fall back to the default cached schema. Can you add that as part of this PR?

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyXSDURIResolver.java
```diff
@@ -60,42 +60,49 @@ public class LibertyXSDURIResolver implements URIResolverExtension, IExternalGra
...
                 }
-                Path serverXSDFile = CacheResourcesManager.getResourceCachePath(SERVER_XSD_RESOURCE);
-                LOGGER.info("Using cached Liberty schema file located at: " + serverXSDFile.toString());
-                return serverXSDFile.toUri().toString();
-            } catch (Exception e) {
-                LOGGER.severe("Error: Unable to deploy server.xsd to lemminx cache.");
-                e.printStackTrace();
             }
+        } catch (Exception e) {
+            LOGGER.severe("Error: Unable to deploy server.xsd to LemMinX cache.");
```
cherylking, line 94: This message is for when generating the schema from the installed Liberty runtime fails right? So I think the log message should be something like "Error: Unable to generate the Liberty schema from the target Liberty runtime.".

cherylking (changes requested):

cherylking (approved):

PR #189, opened 2023-08-01
Generate schema w/ schemaVersion=1.1, outputVersion=2 

### LINKED ISSUE #174: Validation error for `updateTrigger` attribute on the`applicationMonitor` element in server.xml
This error was reported by Danny Haase in IBM Consulting. For the following element/attribute:

```
<applicationMonitor updateTrigger="${env.trigger}" />
```

This diagnostic was displayed:

image

The `updateTrigger` attribute is an enum. So it does not recognize a variable as being valid. Not sure if this should be addressed in the language server, or in the schema itself. It seems that some attributes (like `httpPort` on `httpEndpoint`) that expect a number have the following defined in the schema to allow for variables instead:

```
<xsd:simpleType name="intType">
        <xsd:union memberTypes="xsd:int variableType"/>
    </xsd:simpleType>
```

Can something similar be done for an enum attribute?

scottkurz:
Fixes #174 .

Consider this more of a POC than an actual fix.   I didn't pay much attention to the main vs. 1.x branch or reworking on top of the latest fix in 1.x.   But it at least demonstrates to me the XSD changes desired in #174 would be addressed by merging in these options into our schemaGen invocations.

### DIFF
--- a/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyXSDURIResolver.java
+++ b/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyXSDURIResolver.java
@@ -140,7 +140,8 @@ public class LibertyXSDURIResolver implements URIResolverExtension, IExternalGra
     
                 LOGGER.info("Generating schema file at: " + xsdDestPath);
     
-                ProcessBuilder pb = new ProcessBuilder("java", "-jar", schemaGenJarPath.toAbsolutePath().toString(), xsdDestPath); //Add locale param here
+                ProcessBuilder pb = new ProcessBuilder("java", "-jar", schemaGenJarPath.toAbsolutePath().toString(), "--schemaVersion=1.1",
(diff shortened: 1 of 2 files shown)


scottkurz:
This test fails for me locally..but it fails both running against the HEAD of 'main' and also against my PR

scottkurz:
TODO:  This needs to be updated for the docker devc case

cherylking:
> This test fails for me locally..but it fails both running against the HEAD of 'main' and also against my PR
> 
> > Failed tests:   io.openliberty.LibertyCompletionTest.testXSDElementCompletionItem(): expected: <170> but was: <169>

You need to clear your `.lemminx` cache in your home/user directory on your machine. Then it will work. This happens when we change the default server.xsd to a newer version. In the future, I think I will include the version in the schema file name to avoid this problem. It means more updates to code, but less problems like this one (and for our customers...I think they would not get the new default schema if they had an older one already).

dshimo (approved):

Issue #66, opened 2022-06-10
Support alternate config file/directory location


cherylking:
In the Maven/Gradle plugins, the default configDirectory can be overridden. The various config files (serverXmlFile, serverEnvFile, bootstrapPropertiesFile) can also be overridden. The liberty-plugin-cfg.xml file will contain the values used for those various config files/dir and can be used by our LS to determine if a file is indeed a config file that should have hover/completion/validation support.

This is not part of the MVP. This is included on slide 34 of the UFO.

cherylking:
Although I said we could use the values from the `liberty-plugin-cfg.xml` file, that file will not exist before the Liberty server has been created. The user would get inconsistent behavior based on whether the target/build directory existed for the project.

To address that we would have to potentially parse the build file (pom.xml / build.gradle) to determine the plugin configuration. So far the language server has not needed to parse the build file.

scottkurz:
I also raised the point: could we just look within an opened XML file for a `<server>` root element.   

Cheryl raised the concern that this presents the possibility that a user is getting assistance for an XML file that the LMP/LGP is not configured to use... i.e. a dead/stale file rather than a "live" one.  

Perhaps once the plugin is run a warning could be generated for a case like this... even though the warning wouldn't appear initially?

Another downside though is performance... though parsing with a streaming API like STAX should be quick it still seems like there could be a case where it scales badly.

cherylking:
I have assigned David to lead this, with support from Evie. Some ideas we discussed today for starting on this:

1) First implement using the `liberty-plugin-cfg.xml` to identify and support various additional config files.
2) Look into ways to generate the `liberty-plugin-cfg.xml` without having to run a full build/install, or just some temp file in the `.libertyls` folder that would identify which source files are Liberty config files used in the build file (pom.xml or build.gradle).
3) Parsing the build files in LCLS is absolute last resort.

cherylking:
We discussed this a little bit at today's design issues call. For idea 2 above, if we added a new goal/task to the plugins that simply processed the build file to generate the `liberty-plugin-cfg.xml` file with the various locations of the Liberty config files, then we would need a way to invoke that new goal/task. I proposed that the new goal/task could be invoked at the time that Liberty Tools is determining which projects to add to the Liberty Dashboard. That should also include any manual path for adding projects to the Liberty Dashboard.

scottkurz:
[...]

PR #192, opened 2023-08-03
init stax component

dshimo:
partial address #66 

- new StAX reader class XmlReader
  - hasServerRoot, readLibertyPluginConfigXml
- a few very basic tests and sample files

remaining items:
-  incorporate new isConfigXmlFile with heuristic, and remove unnecessary pieces


lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/XmlReader.java
```diff
@@ -0,0 +1,89 @@
...
+import javax.xml.stream.XMLEventReader;
+import javax.xml.stream.XMLInputFactory;
+import javax.xml.stream.XMLStreamException;
+import javax.xml.stream.events.XMLEvent;
+
+public class XmlReader {
+    private static final Logger LOGGER = Logger.getLogger(XmlReader.class.getName());
+    private static XmlReader instance;
+
+    public static XmlReader getInstance() {
```
cherylking, line 17: Why is this a singleton? Couldn't we just have this be a utility class with static methods? Just wondering what the reasons were for it being a singleton.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/XmlReader.java
```diff
@@ -0,0 +1,89 @@
+package io.openliberty.tools.langserver.lemminx.services;
```
cherylking, line 1: add copyright

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/XmlReader.java
```diff
@@ -0,0 +1,89 @@
...
+            e.printStackTrace();
+        } catch (XMLStreamException e) {
+            LOGGER.severe("Failed to grab the first starting element for file: " + file);
+            e.printStackTrace();
+        }
+
+        return false;
+    }
+
+    public void readLibertPluginConfigXml(File file) {
```
cherylking, line 48: I would like this to be more flexible, but I can rework it in my own PR.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyUtils.java
```diff
@@ -73,6 +74,9 @@ public static boolean isConfigXMLFile(String filePath) {
 
         return isServerXMLFile(filePath) || isConfigDirFile(filePath) ||
                 LibertyProjectsManager.getInstance().getWorkspaceFolder(filePath).hasConfigFile(filePath);
+
```
cherylking, line 77: just a comment for going forward...I think we may still want to automatically consider any file that returns true for `isConfigDirFile` or the `src/main/liberty/config/server.xml` as a Liberty config file and apply our xsd to it...even if it does not have a root element of `server`...because it would cause a failure in Liberty server startup if it didn't have a `server` root element.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/XmlReader.java
```diff
@@ -0,0 +1,89 @@
...
+        try {
+            XMLEventReader reader = factory.createXMLEventReader(new FileInputStream(file));
+            if (reader.hasNext()) {
+                XMLEvent firstTag = reader.nextTag(); // first start/end element
+                reader.close();
+                return isServerElement(firstTag);
+            }
+            reader.close();
+        } catch (FileNotFoundException e) {
+            e.printStackTrace();
```
cherylking, line 39: probably should log something here...although it should never happen

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/XmlReader.java
```diff
@@ -0,0 +1,89 @@
...
+
+        XMLInputFactory factory = XMLInputFactory.newInstance();
+        try {
+            XMLEventReader reader = factory.createXMLEventReader(new FileInputStream(file));
+            if (reader.hasNext()) {
+                XMLEvent firstTag = reader.nextTag(); // first start/end element
+                reader.close();
+                return isServerElement(firstTag);
+            }
+            reader.close();
```
cherylking, line 37: maybe define `reader` outside of the try/catch and then close it in a `finally` if it is not null

cherylking (changes requested):

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyUtils.java
dshimo: yes agreed. in a local change, I made the isConfigXmlFile logic to be 
```
return isServerXMLFile(filePath) || isConfigDirFile(filePath) || XmlReader.hasServerRoot(new File(filePath))
```

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/XmlReader.java
```diff
@@ -0,0 +1,147 @@
...
+            return false;
+        }
+        XMLInputFactory factory = XMLInputFactory.newInstance();
+        XMLEventReader reader = factory.createXMLEventReader(new FileInputStream(file));
+        if (reader.hasNext()) {
+            XMLEvent firstTag = reader.nextTag(); // first start/end element
+            reader.close();
+            return isServerElement(firstTag);
+        }
+        reader.close();
```
cherylking, line 40: So if something in this method throws an exception, won't the reader remain open? You should add a try/finally.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/XmlReader.java
```diff
@@ -26,44 +26,75 @@
...
         }
+
         XMLInputFactory factory = XMLInputFactory.newInstance();
-        XMLEventReader reader = factory.createXMLEventReader(new FileInputStream(file));
-        if (reader.hasNext()) {
-            XMLEvent firstTag = reader.nextTag(); // first start/end element
-            reader.close();
-            return isServerElement(firstTag);
+        XMLEventReader reader = null;
+        try {
```
cherylking, line 36: You slacked me and I responded, but guess you did not see. You can do a try/finally without a catch, so that you can keep the model of throwing the exceptions on the method signature.

cherylking (approved):

PR #194, opened 2023-08-08
change config file logic

dshimo:
addresses *.xml portion of #66
uses server heuristic to support server config files anywhere.
simplifies tracking and diagnostics continue to work


lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/XmlReader.java
```diff
@@ -30,24 +32,30 @@
...
-            return false;
-        }
-
+    public static boolean hasServerRoot(String filePath) {
         XMLInputFactory factory = XMLInputFactory.newInstance();
         XMLEventReader reader = null;
+        File file = null;
+        
         try {
+            file = new File(new URI(filePath).getPath());
```
cherylking, line 41: just curious why we have to create a URI and get the path from that to create the file? What is the format of the filePath? I know we changed `\` to `/` in the calling method.

cherylking (commented):
Just the one question

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/XmlReader.java
dshimo: I learned checking `file.exists()` with a `File` constructed with a string in the URI format returns `false`. We get URI formatting from using DOMDocument

lemminx-liberty/src/test/java/io/openliberty/LibertyDiagnosticTest.java
```diff
@@ -141,15 +141,5 @@ public void testDiagnosticsForInclude() throws IOException {
...
-
-        Collection<LibertyWorkspace> workspaceFolders = LibertyProjectsManager.getInstance().getLibertyWorkspaceFolders();
-        assertTrue(workspaceFolders.size() == 1);
-
-        LibertyWorkspace libWorkspace = workspaceFolders.iterator().next();
-
-        assertTrue(libWorkspace.hasConfigFile(new File("src/test/resources/empty_server.xml").getCanonicalPath()));
-        assertFalse(libWorkspace.hasConfigFile("MISSING FILE"));
-        assertFalse(libWorkspace.hasConfigFile("MULTI LINER"));
-        assertFalse(libWorkspace.hasConfigFile("MISSING FILE.xml"));
```
cherylking, line 153: Should we test that `isConfigXMLFile` works correctly for some of these files?

cherylking (changes requested):

cherylking (approved):

Issue #188, opened 2023-08-01
For project using Gradle and Maven both, the Gradle install is detected as newer even with the Maven project has been more recently used.


scottkurz:
If I do:
> $ gradle libertyDev

then 
> $ mvn liberty:dev

in the same project, the LCLS logic is going to have to guess which install the server.xml is associated with.

It looks like [here](https://github.com/OpenLiberty/liberty-language-server/blob/dbf22fa124ff83d9cb0634070273345221b5ba0e/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyUtils.java#L120-L138) we attempt to use the lastModified timestamp, e.g. for the 'openliberty.properties' file.

However, I'm just observing on my local Windows install that in this case, the Gradle install always shows a newer timestamp for this file, even the Maven install was done more recently.   The timestamp seems to be preserved in the Maven artifact archive, whereas in the Gradle case it is set during install when the file is copied into place.

E.g. I have this although I ran the `mvn liberty:dev` more recently than the gradle one, for OL 23.0.0.7:

> -rw-r--r-- 1 GMX+854188897 4096 922 Aug  1 14:11 ./build/wlp/lib/versions/openliberty.properties
> -rw-r--r-- 1 GMX+854188897 4096 922 Jul 10 15:05 ./target/liberty/wlp/lib/versions/openliberty.properties

scottkurz:
Not sure if this is a good time to ask the question:  but I wonder how a path like this works if the install dir is external to the build/target dir altogether?    (Don't feel obliged to answer if it's a totally separate path).

cherylking:
So my thoughts on this are we could traverse the path up to the dir containing `wlp` to compare timestamps. 

As for an external install dir, we do not currently handle that at all. It would use the default server schema because the code would not find a `ws-schemagen.jar` in the workspace. This is a totally separate path that will potentially involve needing to build the project (or invoke some setup goal/task at a minimum) when it is imported into the Liberty dashboard.

PR #195, opened 2023-08-08
Rework loading of Liberty runtime info

### LINKED ISSUE #191: Handle external installation directory correctly
If the Liberty installation is located outside of the target/build directory, some functionality is lacking. For example, the schema cannot be generated to validate the server.xml causing the default server schema to be used. This is because the code looks for the `ws-schemagen.jar` in the target/build directory and does not find it.

This scenario occurs when the `installDirectory` parameter is specified for the Liberty Maven plugin, or the `installDir` property is specified for the Liberty Gradle plugin. Once Liberty is installed, the location of the installation can be determined from the `installDirectory` element in the `liberty-plugin-config.xml` file which is in the target/build directory.

```
    <installDirectory>/path/to/liberty/wlp</installDirectory>
```

So if that `installDirectory` path does not start with the Liberty workspace path in the IDE, we should look for the `ws-schemagen.jar` at that path instead.

This would also affect feature list generation, which uses the `ws-featurelist.jar` from Liberty.

Finally, the basic determination of which runtime (OpenLiberty vs WebSphere Liberty) and version are installed (getRuntimeInfo() and getVersion() in LibertyUtils.java) does not work correctly in this scenario.

Need to determine if there is any other functionality lacking when the Liberty installation is not in the Liberty workspace for the IDE.

cherylking:
Fixes #191 

In addition to supporting/handling external Liberty installation locations, the loading of the runtime info has been streamlined. Now the runtime type, version and location are all determined in one method call.

I also changed the lookup methods for Liberty jar files (ws-schemagen.jar and ws-featurelist.jar) to first check for the `liberty-plugin-config.xml` and use the `installDirectory` to determine where to look for the jar file. If the `liberty-plugin-config.xml` is not found, then it falls back to simply look in the LibertyWorkspace for the jar file.


cherylking:
Trying to figure out how/why a `ol-22.0.0.12.xsd` file is getting generated or copied to the `modC` module. The correct `ol-23.0.0.6.xsd` is generated and used. So the test case passes. But I cannot explain the presence of the other xsd.

Screenshot 2023-08-09 at 5 57 21 PM

cherylking:
@scottkurz This may also fix #188, but would need you to test once I create a snapshot. The new code defaults to locating the `wsschema-gen.jar` using the `installDirectory` obtained from the `liberty-plugin-config.xml` file. In your scenario, it should find two of those `liberty-plugin-config.xml` files and use the one with the latest modified timestamp which should be the correct installation to use.

 If that file is not found, then it uses the old behavior of searching for it in the LibertyWorkspace which can give the wrong jar based on timestamps of the actual Liberty installation.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyHoverParticipant.java
```diff
@@ -62,17 +64,16 @@ public Hover onText(IHoverRequest request, CancelChecker cancelChecker) {
 	}
 
 	private Hover getHoverFeatureDescription(String featureName, DOMDocument domDocument) {
```
dshimo, line 66: The indentation is uneven here

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyHoverParticipant.java
```diff
@@ -62,17 +64,16 @@ public Hover onText(IHoverRequest request, CancelChecker cancelChecker) {
...
-            String libertyRuntimeVersionInfo = LibertyUtils.getRuntimeAndVersionInfo(domDocument);
-            String libertyVersion =  LibertyUtils.getVersionFromInfo(libertyRuntimeVersionInfo);
-            String libertyRuntime =  LibertyUtils.getRuntimeFromInfo(libertyRuntimeVersionInfo);
-
+        LibertyRuntime runtimeInfo = LibertyUtils.getLibertyRuntimeInfo(domDocument);
```
dshimo, line 67: The indentation is uneven in this method

dshimo (approved):

PR #206, opened 2023-08-25
Post process schema

### LINKED ISSUE #186: server.xml assist does not provide diagnostics for incorrect attributes
### USE CASE
As one example, 

take the fact that we have different "context root" attribute names e.g. we have:
```xml
<webApplication contextRoot="ctx"
```
but
```xml
<web-ext context-root="ctx"..
```

It would be nice to have LCLS validate this and provide a diagnostic at least (not sure a quick fix is possible, but the user could do Ctrl+Space to get a completion suggestion to fix easily enough themselves).

### PROBLEM 

However, the Liberty XSD generated is something like:
```xml
    <xsd:complexType name="com.ibm.ws.app.manager.webappcfg"> 
        ...
        <xsd:anyAttribute processContents="skip"/>
```
preventing the XSD from detecting this as invalid.

### RUNTIME ISSUE LINK
This depends on https://github.com/OpenLiberty/open-liberty/issues/25579.   Presumably we'd address in the runtime and then see what if anything would be needed to pick up the change in LCLS.

cherylking:
Fixes #186 

This PR updates the default cached schema to remove the anyAttribute from elements without extraProperties. It also does the same post processing on generated schema files for local Liberty installations and Liberty running in containers.

Here is the before (no diagnostic on the `bogus` attribute):

Screenshot 2023-08-25 at 2 01 22 PM

And the after:

Screenshot 2023-08-25 at 2 07 06 PM


cherylking:
Since git won't show the diff for the xsd file, I am including a couple screenshots of a diff in VS Code. You can see it is a lot of formatting differences and explicit end elements instead of `/>` on the element itself. The second screenshot shows an example of the `anyAttribute` removal.

Screenshot 2023-08-25 at 3 03 59 PM

Screenshot 2023-08-25 at 3 06 48 PM

dshimo (approved):

PR #219, opened 2023-10-05
Handle Podman container type in devc metadata file

### LINKED ISSUE #199: Handle podman as an alternative container for devc
Once podman is supported by the Liberty Maven and Gradle plugins, need to add support to detect the type of container for determining the Liberty runtime and version installed.

This is relevant in the DockerServer.java class.

mattbsox:
Fixes #199

### DIFF
--- a/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/LibertyWorkspace.java
+++ b/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/LibertyWorkspace.java
@@ -165,9 +165,7 @@ public class LibertyWorkspace {
     }
 
     public void setContainerType(String containerType) {
-        if (containerType != null && !containerType.isEmpty()) {
-            this.containerType = containerType;
-        }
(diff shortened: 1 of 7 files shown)


lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/LibertyWorkspace.java
```diff
@@ -158,6 +160,16 @@ public void setContainerName(String containerName) {
...
     }
 
+    public String getContainerType() {
+        return containerType;
+    }
+
+    public void setContainerType(String containerType) {
+        if (containerType != null && !containerType.isEmpty()) {
+            this.containerType = containerType;
+        }
```
cherylking, line 170: If the containerType was set to podman, and then subsequently null was passed in here because they used an older version of the plugin, we should default back to docker. I think there should be an ` else { this.containerType = "docker"; }`.

cherylking (changes requested):

cherylking (approved):

Issue #67, opened 2022-06-10
Provide feature suggestion when config element is for feature that is not configured


cherylking:
Provide feature suggestions when config element is used but the feature is not currently configured. WDT does this by generating a featureList.xml file which maps config elements to their features.

This is a future enhancement and not part of the MVP. Slide 36 of UFO.

cherylking:
Implementation notes:

1. Generate feature list for installed runtime. Process the feature list to gather data.
2. Create a map where the key is the config element and the entry is a collection of feature names that utilize that config element.
3. Validate that each config element in a Liberty configuration file is supported by a feature in the featureManager list.
4. Issue a diagnostic for any config element without a related installed feature, and a quick fix with a list of possible features to add that support that element.

Question: Do we need to consider included files and the features they add, as well as the configDropins files (defaults and overrides)? Can first pass of support just consider the features within the same file?

cherylking:
Side note: Could also collect the list of features included by each feature when processing the feature list. This is needed to support issue #164.

cherylking:
Now that this is merged, the only piece not handled is the big picture of all features that are configured across all Liberty configuration files (server.xml can include other xml files, and configDropins can contain config snippets). I will open a separate issue for this.

PR #220, opened 2023-10-05
Diagnostics and Quick fixes for lost config elements

dshimo:
Addresses #67 

CodeActionParticipant, AddFeature:
- Add an initial draft for 'AddFeature' quickfix.
- The idea for the quickfix is to provide a list of direct features because transitive relations offer too many.
- As an enhancement, this quick fix could alternatively provide a list of some of the most commonly used features.

DiagnosticParticipant:
- Adds additional check while parsing the `server.xml` to recognize config elements.
- After the document is parsed, create relevant diagnostics for lost config elements.
- Add a trigger line for grabbing the feature list even if a FeatureManager is empty or not found

**New** FeatureListGraph class:
- Designed as a manager class for retrieving a desired FeatureListNode
- On assumption that all features and config elements have a unique identifier, this class uses a hashmap to provide constant time lookup to a node.
- To provide a list of transitive relations for features -> config element or vice versa, FeatureListGraph implements a breadth-first-search iterative strategy on a given node.

**New** FeatureListNode class:
- A new class to represent a generic node in a Feature List XML. It's intended to store features or config elements.
- isConfigElement operates on the heuristic that features provide major versions and that `.` isn't typical in a normal element name.
- 'Enables' refers to when this node automatically pulls in another feature or config element.
- 'Enablers' refers to features that would automatically pull in this node.

FeatureService:
- The priority for using a generated local feature list is placed higher. This xml file not only provides the list of features but also the necessary metadata for recognizing config elements and their relations.
- Build the FeatureListGraph while reading the feature list

Tests:
- Add check that diagnostics appear on expected locations
- Add check that the feature list xml gets digested within expectations

### DIFF
--- a/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
+++ b/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
@@ -26,6 +26,8 @@ import org.eclipse.lsp4j.jsonrpc.CancelChecker;
 import io.openliberty.tools.langserver.lemminx.data.FeatureListGraph;
 import io.openliberty.tools.langserver.lemminx.data.LibertyRuntime;
 import io.openliberty.tools.langserver.lemminx.services.FeatureService;
+import io.openliberty.tools.langserver.lemminx.services.LibertyProjectsManager;
+import io.openliberty.tools.langserver.lemminx.services.LibertyWorkspace;
 import io.openliberty.tools.langserver.lemminx.services.SettingsService;
 import io.openliberty.tools.langserver.lemminx.util.*;
 
@@ -55,7 +57,7 @@ public class LibertyDiagnosticParticipant implements IDiagnosticsParticipant {
 
     public static final String INCORRECT_FEATURE_CODE = "incorrect_feature";
 
-    private Set<String> includedFeatures = null;
+    private Set<String> includedFeatures;
     
     @Override
     public void doDiagnostics(DOMDocument domDocument, List<Diagnostic> diagnostics,
@@ -73,16 +75,10 @@ public class LibertyDiagnosticParticipant implements IDiagnosticsParticipant {
     private void validateDom(DOMDocument domDocument, List<Diagnostic> diagnosticsList) throws IOException {
         List<DOMNode> nodes = domDocument.getDocumentElement().getChildren();
         List<Diagnostic> tempDiagnosticsList = new ArrayList<Diagnostic>();
-
-        FeatureListGraph featureGraph = FeatureService.getInstance().getFeatureListGraph();
+        includedFeatures = new HashSet<>();
+        LibertyWorkspace workspace = LibertyProjectsManager.getInstance().getWorkspaceFolder(domDocument.getDocumentURI());
         // TODO: Consider adding a cached feature list onto repo to optimize
-        if (featureGraph.isEmpty()) {
-            LibertyRuntime runtimeInfo = LibertyUtils.getLibertyRuntimeInfo(domDocument);
-            String libertyVersion =  runtimeInfo == null ? null : runtimeInfo.getRuntimeVersion();
-            String libertyRuntime =  runtimeInfo == null ? null : runtimeInfo.getRuntimeType();
-            FeatureService.getInstance().getInstalledFeaturesList(domDocument.getDocumentURI(), libertyRuntime, libertyVersion);
-        }
-
+        FeatureListGraph featureGraph = workspace.getFeatureListGraph();
         for (DOMNode node : nodes) {
             String nodeName = node.getNodeName();
             if (LibertyConstants.FEATURE_MANAGER_ELEMENT.equals(nodeName)) {
@@ -105,7 +101,6 @@ public class LibertyDiagnosticParticipant implements IDiagnosticsParticipant {
 
         // Search for duplicate features
         // or features that do not exist
-        Set<String> includedFeatures = new HashSet<>();
         List<DOMNode> features = featureManager.getChildren();
         for (DOMNode featureNode : features) {
             DOMNode featureTextNode = (DOMNode) featureNode.getChildNodes().item(0);
@@ -131,7 +126,6 @@ public class LibertyDiagnosticParticipant implements IDiagnosticsParticipant {
                 }
             }
         }
-        this.includedFeatures = includedFeatures;
     }
 
     /**
@@ -191,7 +185,8 @@ public class LibertyDiagnosticParticipant implements IDiagnosticsParticipant {
      */
     private void holdConfigElement(DOMDocument domDocument, DOMNode configElementNode, List<Diagnostic> tempDiagnosticsList) {
         String configElementName = configElementNode.getNodeName();
-        Range range = XMLPositionUtility.createRange(configElementNode.getStart(), configElementNode.getEnd(), domDocument);
+        Range range = XMLPositionUtility
+                .createRange(configElementNode.getStart(), configElementNode.getEnd(), domDocument);
         Diagnostic tempDiagnostic = new Diagnostic(range, MISSING_CONFIGURED_FEATURE_MESSAGE, null, LIBERTY_LEMMINX_SOURCE, MISSING_CONFIGURED_FEATURE_CODE);
         tempDiagnostic.setSource(configElementName);
         tempDiagnosticsList.add(tempDiagnostic);
@@ -205,7 +200,7 @@ public class LibertyDiagnosticParticipant implements IDiagnosticsParticipant {
      */
     private void validateConfigElements(List<Diagnostic> diagnosticsList, List<Diagnostic> tempDiagnosticsList, FeatureListGraph featureGraph) {
         for (Diagnostic tempDiagnostic : tempDiagnosticsList) {
-            if (includedFeatures == null) {
+            if (includedFeatures.isEmpty()) {
                 diagnosticsList.add(tempDiagnostic);
                 continue;
             }
--- a/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
+++ b/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
@@ -76,8 +76,13 @@ public class LibertyDiagnosticParticipant implements IDiagnosticsParticipant {
         List<DOMNode> nodes = domDocument.getDocumentElement().getChildren();
         List<Diagnostic> tempDiagnosticsList = new ArrayList<Diagnostic>();
 
-        // TODO: consider initiallizing FeatureService even when features aren't detected
         FeatureListGraph featureGraph = FeatureService.getInstance().getFeatureListGraph();
+        if (featureGraph.isEmpty()) {
+            LibertyRuntime runtimeInfo = LibertyUtils.getLibertyRuntimeInfo(domDocument);
+            String libertyVersion =  runtimeInfo == null ? null : runtimeInfo.getRuntimeVersion();
+            String libertyRuntime =  runtimeInfo == null ? null : runtimeInfo.getRuntimeType();
+            FeatureService.getInstance().getInstalledFeaturesList(domDocument.getDocumentURI(), libertyRuntime, libertyVersion);
+        }
 
         for (DOMNode node : nodes) {
             String nodeName = node.getNodeName();
@@ -86,11 +91,9 @@ public class LibertyDiagnosticParticipant implements IDiagnosticsParticipant {
             } else if (LibertyConstants.INCLUDE_ELEMENT.equals(nodeName)) {
                 validateIncludeLocation(domDocument, diagnosticsList, node);
             } else if (featureGraph.isConfigElement(nodeName)) {
-                LOGGER.warning("Detected a config element!");
                 holdConfigElement(domDocument, diagnosticsList, node, tempDiagnosticsList);
             }
         }
-        LOGGER.warning("Entering validateConfigElements");
         validateConfigElements(diagnosticsList, tempDiagnosticsList, featureGraph);
     }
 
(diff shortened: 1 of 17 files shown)


lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/codeactions/AddFeature.java
```diff
@@ -0,0 +1,79 @@
...
+                insertText = "\n\t<feature>microProfile-5.0</feature>";
+            }
+
+            String configElement = diagnostic.getSource();
+            if (configElement.equals("ssl")) {
+                FeatureService.getInstance().getFeatureListGraph().get("ssl").getEnablers();
+            }
+            Set<String> featureCandidates = FeatureService.getInstance().getFeatureListGraph().get(configElement).getEnablers();
+            codeActions.add(CodeActionFactory.insert("This is a test: " + featureCandidates, insertPosition, insertText, document.getTextDocument(), diagnostic));
+            for (String feature : featureCandidates) {
```
dshimo, line 70: VS Code seems to have a limit of about 10 or 12 quickfixes to a diagnostic.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/FeatureService.java
```diff
@@ -72,8 +73,11 @@ public static FeatureService getInstance() {
     private List<Feature> defaultFeatureList;
     private long featureUpdateTime;
 
+    private FeatureListGraph featureListGraph;
+
     private FeatureService() {
         featureCache = new HashMap<>();
+        featureListGraph = new FeatureListGraph();
```
cherylking, line 80: Does this need to be a Map just like `featureCache`? This graph should be specific to the Liberty installation. If they have multiple projects with different runtimes installed, need to use the correct `FeatureListGraph`. Even caching by `ol-23.0.0.6` may not be sufficient in the long run, if they have two different projects using that runtime with a different set of features installed. Maybe the graph should be stored in the `LibertyWorkspace`?

cherylking (changes requested):
My comment about the FeatureListGraph in FeatureService is significant and requires some code changes. Hopefully it's doable.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
```diff
@@ -191,10 +197,16 @@ private void holdConfigElement(DOMDocument domDocument, List<Diagnostic> diagnos
...
+     * Compare the required feature set with included feature set for each config element.
+     * @param diagnosticsList
+     * @param tempDiagnosticsList
+     * @param featureGraph
+     */
     private void validateConfigElements(List<Diagnostic> diagnosticsList, List<Diagnostic> tempDiagnosticsList, FeatureListGraph featureGraph) {
         for (Diagnostic tempDiagnostic : tempDiagnosticsList) {
             String configElement = tempDiagnostic.getSource();
-            Set<String> includedFeaturesCopy = new HashSet<String>(includedFeatures);
+            Set<String> includedFeaturesCopy = (includedFeatures == null) ? new HashSet<String>() : new HashSet<String>(includedFeatures);
```
cherylking, line 209: If `includedFeatures` is null, then there is no point in doing the following 5 lines. The `includedFeaturesCopy` will always be empty in that case.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/codeactions/AddFeature.java
```diff
@@ -20,75 +20,62 @@
...
+                insertText = FEATUREMANAGER_FORMAT;
+            }
 
-            // XMLGenerator generator = request.getXMLGenerator();
-            // if (!request.canSupportResolve()) {
-            //     insertStrRequired = generator.generateMissingElements(request.getComponent(ContentModelManager.class), document.getDocumentElement(), true);
-            //     insertStrAll = generator.generateMissingElements(request.getComponent(ContentModelManager.class), document.getDocumentElement(), false);
-            // }
+            Range nodeRange = XMLPositionUtility.createRange(insertNode.getStart(), insertEndOffset, document);
+            Logger.getLogger("dshi").warning("startCharacter" + nodeRange.getStart().getCharacter());
```
cherylking, line 69: `getLogger("dshi")`???

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
```diff
@@ -53,22 +65,36 @@ public void doDiagnostics(DOMDocument domDocument, List<Diagnostic> diagnostics,
...
-            if (LibertyConstants.FEATURE_MANAGER_ELEMENT.equals(node.getNodeName())) {
-                validateFeature(domDocument, list, node);
-            } else if (LibertyConstants.INCLUDE_ELEMENT.equals(node.getNodeName())) {
-                validateIncludeLocation(domDocument, list, node);
+            String nodeName = node.getNodeName();
+            if (LibertyConstants.FEATURE_MANAGER_ELEMENT.equals(nodeName)) {
+                validateFeature(domDocument, diagnosticsList, node);
+            } else if (LibertyConstants.INCLUDE_ELEMENT.equals(nodeName)) {
+                validateIncludeLocation(domDocument, diagnosticsList, node);
+            } else if (featureGraph.isConfigElement(nodeName)) {
```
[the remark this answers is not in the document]
dshimo: `<feature>` elements are children of the `<featureManager>` and therefore not included in the for-each loop that iterates over the children of the `server` root.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/codeactions/AddFeature.java
dshimo: AAAH
I will triple check what I check in from now on.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/codeactions/AddFeature.java
```diff
@@ -20,75 +20,62 @@
...
-                    insertText += FEATURE_FORMAT;
+                    insertText = (insertNode == null) ? FEATURE_FORMAT.replace("\n", "\n\t") : FEATURE_FORMAT;
+                    insertNode = (insertNode == null) ? node : insertNode;
                     break;
                 }
             }
 
-            // String insertStrRequired = null;
-            // String insertStrAll = null;
+            int insertEndOffset = (insertNode == null) ? ((DOMElement)document.getDocumentElement()).getStartTagCloseOffset()+1 : insertNode.getEnd();
```
evie-lau, line 60: Maybe add a comment here explaining the end of `<server>`/`<featureManager> tag vs end of previously found `</feature>` tag?

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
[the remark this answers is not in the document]
cherylking: Ah..ok. I didn't realize it was just the direct children of the `<server>` element, but that makes sense. Thanks.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
```diff
@@ -53,22 +67,29 @@ public void doDiagnostics(DOMDocument domDocument, List<Diagnostic> diagnostics,
...
         }
     }
 
-    private void validateDom(DOMDocument domDocument, List<Diagnostic> list) throws IOException {
+    private void validateDom(DOMDocument domDocument, List<Diagnostic> diagnosticsList) throws IOException {
         List<DOMNode> nodes = domDocument.getDocumentElement().getChildren();
-
+        List<Diagnostic> tempDiagnosticsList = new ArrayList<Diagnostic>();
+        includedFeatures = new HashSet<>();
+        LibertyWorkspace workspace = LibertyProjectsManager.getInstance().getWorkspaceFolder(domDocument.getDocumentURI());
```
cherylking, line 79: Is it possible for this to return null?

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
```diff
@@ -144,15 +164,53 @@ private void validateIncludeLocation(DOMDocument domDocument, List<Diagnostic> l
...
+     * Compare the required feature set with included feature set for each config element.
+     * @param diagnosticsList
+     * @param tempDiagnosticsList
+     * @param featureGraph
+     */
+    private void validateConfigElements(List<Diagnostic> diagnosticsList, List<Diagnostic> tempDiagnosticsList, FeatureListGraph featureGraph) {
+        for (Diagnostic tempDiagnostic : tempDiagnosticsList) {
+            if (includedFeatures.isEmpty()) {
+                diagnosticsList.add(tempDiagnostic);
+                continue;
```
cherylking, line 205: We can actually optimize this. If `includedFeatures.isEmpty()`, it will be for every iteration. Just check before entering the `for` loop and `addAll` of the temp diagnostics to the list and return.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/codeactions/AddFeature.java
```diff
@@ -0,0 +1,108 @@
...
+            referenceRangeStart = server.getStart();
+            referenceRangeEnd = server.getStartTagCloseOffset()+1;
+        }
+        Range referenceRange = XMLPositionUtility.createRange(referenceRangeStart, referenceRangeEnd, document);
+
+        try {
+            String indent = request.getXMLGenerator().getWhitespacesIndent();
+            insertText = IndentUtil.formatText(insertText, indent, referenceRange.getStart().getCharacter());
+        } catch (BadLocationException e) {
+            e.printStackTrace();
```
cherylking, line 94: Shouldn't `printStackTrace()` in the code we merge. Is it possible to log messages in this class?

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/codeactions/AddFeature.java
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        // getAllEnabledBy would return all transitive features but typically offers too much
+        Set<String> featureCandidates = LibertyProjectsManager.getInstance()
```
cherylking, line 98: Is it possible for this Set to be empty? Just wondering if we should make sure it is non-empty before we spend all the effort of figuring out the text to enter for the quick fix above?

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/FeatureService.java
```diff
@@ -374,21 +377,40 @@ public List<Feature> readFeaturesFromFeatureListFile(List<Feature> installedFeat
...
+                        configNode.addEnabledBy(currentFeature);
+                        currentFeatureNode.addEnables(configElement);
+                    }
+                }
             }
             installedFeatures = featureInfo.getFeatures();
+            libertyWorkspace.setFeatureListGraph(featureListGraph);
             libertyWorkspace.setInstalledFeatureList(installedFeatures);
         } else {
             LOGGER.warning("Unable to get installed features for current Liberty workspace: " + libertyWorkspace.getWorkspaceString());
```
cherylking, line 412: Should the featureListGraph in the libertyWorkspace get cleared out in this case?

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/LibertyWorkspace.java
```diff
@@ -206,4 +210,24 @@ public String toString() {
...
+        this.featureListGraph = featureListGraph;
+        if (isLibertyInstalled) {
+            this.featureListGraph.setRuntime(libertyRuntime + "-" + libertyVersion);
+        }
+    }
+
+    public FeatureListGraph getFeatureListGraph() {
+        String workspaceRuntime = libertyRuntime + "-" + libertyVersion;
+        boolean generateGraph = featureListGraph.isEmpty() || !featureListGraph.getRuntime().equals(workspaceRuntime);
+        if (this.isLibertyInstalled && generateGraph) {
```
cherylking, line 223: There is a method in this class that clears out the installed feature list when liberty is not installed. That method should also clear out the feature list graph.

cherylking (changes requested):

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/codeactions/AddFeature.java
```diff
@@ -87,17 +103,13 @@ public void doCodeAction(ICodeActionRequest request, List<CodeAction> codeAction
...
         Range referenceRange = XMLPositionUtility.createRange(referenceRangeStart, referenceRangeEnd, document);
 
+        String indent = "    ";
         try {
-            String indent = request.getXMLGenerator().getWhitespacesIndent();
-            insertText = IndentUtil.formatText(insertText, indent, referenceRange.getStart().getCharacter());
+            indent = request.getXMLGenerator().getWhitespacesIndent();
         } catch (BadLocationException e) {
-            e.printStackTrace();
+            Logger.getLogger(AddFeature.class.getName()).info("Defaulting indent to four spaces.");
```
cherylking, line 110: Replace with LOGGER variable

cherylking (commented):
See one comment about LOGGER variable.

cherylking (approved):

PR #226, opened 2023-10-11
Use default feature list for config element validation

cherylking:


lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/codeactions/AddFeature.java
```diff
@@ -57,14 +63,30 @@ public void doCodeAction(ICodeActionRequest request, List<CodeAction> codeAction
...
         TextDocument textDocument = document.getTextDocument();
+
+        LibertyWorkspace ws = LibertyProjectsManager.getInstance().getWorkspaceFolder(document.getDocumentURI());    
+        if (ws == null) {
+            LOGGER.warning("Could not add quick fix for missing feature because could not find Liberty workspace for document: "+document.getDocumentURI());
+            return;
+        }
+
+        FeatureListNode flNode = ws.getFeatureListGraph().get(diagnostic.getSource());
+        if (flNode == null) {
```
dshimo, line 74: I can't help but think that the graph wouldn't return a null node in the quick fix if the diagnostic was made. The null safeguard seems proper though.

dshimo (approved):

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/codeactions/AddFeature.java
cherylking: I think it would probably only happen in the test path.

PR #232, opened 2023-10-18
Fix hover unexpectedly showing, Support configDirectory

evie-lau:
- Fix hover showing up in all .env and .properties files when they aren't defined as custom configs
- Add support for configDirectory default config files
- Cleanup output logs


liberty-ls/src/main/java/io/openliberty/tools/langserver/LibertyConfigFileManager.java
```diff
@@ -109,6 +114,12 @@ public static boolean isServerEnvFile(LibertyTextDocument tdi) {
      */
     public static boolean isServerEnvFile(String uri) {
         String filePath = normalizeFilePath(uri);
+        for (String configDir : customConfigDirs) {
```
cherylking, line 117: I think we should check the conditions that are straight forward first (that are part of the return statement), then if that isn't true...check the config dirs.

Also, why not strip the filename off (after checking that it is `server.env`) and check if the customConfigDirs contain a match?

cherylking (changes requested):

liberty-ls/src/main/java/io/openliberty/tools/langserver/LibertyConfigFileManager.java
evie-lau: Much cleaner with that method change

cherylking (approved):

cherylking (approved):

TrevCraw (approved):

Issue #230, opened 2023-10-12
Consider all features enabled by Liberty configuration files for diagnostic checks


cherylking:
Follow on to issue #67. The diagnostic code for config elements that do not have a feature configured only takes into account the features listed in the current xml file. If other xml files are included via the `<include>` element, or by placing them in the `configDropins` folder, any features listed in those xml files are not considered in the diagnostic check.

cherylking:
This also means that config snippets in `configDropins` will show diagnostics even though the feature is configured in the xml file that includes it.

cherylking:
Decided to put this one on pause until we get customer feedback.

PR #233, opened 2023-10-23
Disable config element validation when no featureManager present

cherylking:
Related to #230 

Until we address 230, temporarily disable the config element validation for missing features when no `<featureManager>` element is found in the current xml file.


evie-lau (approved):

scottkurz (approved):
LGTM

PR #235, opened 2023-10-25
Support all bootstrap.properties and server.env files

### LINKED ISSUE #234: Allow LS support for all `bootstrap.properties` and `server.env` file
As per discussion in https://github.com/OpenLiberty/liberty-language-server/issues/213#issuecomment-1779326611, to align with providing liberty-lemminx support for all XML files with a <server> root, we should also allow LCLS support for all `bootstrap.properties` and `server.env` files.

This modifies some of the original goals of #66 , and takes precedence over the original design

evie-lau:
Fixes #234 

- Provide LCLS support for all bootstrap.properties and server.env files, instead of limiting to `src/main/liberty/config` and specified `configDirectory`. This better aligns with how we support all XML files with `<server>` root for lemminx-liberty.

### DIFF
--- a/liberty-ls/src/main/java/io/openliberty/tools/langserver/LibertyConfigFileManager.java
+++ b/liberty-ls/src/main/java/io/openliberty/tools/langserver/LibertyConfigFileManager.java
@@ -36,8 +36,8 @@ public class LibertyConfigFileManager {
     public static final String CUSTOM_BOOTSTRAP_PROPERTIES_XML_TAG = "bootstrapPropertiesFile";
     public static final String CUSTOM_SERVER_CONFIG_DIR_TAG = "configDirectory";
 
-    private static final String DEFAULT_SERVER_ENV = "src/main/liberty/config/server.env".replace("/", File.separator);
-    private static final String DEFAULT_BOOTSTRAP_PROPERTIES = "src/main/liberty/config/bootstrap.properties".replace("/", File.separator);
+    public static final String SERVER_ENV_FILENAME = "server.env";
+    public static final String BOOTSTRAP_PROPERTIES_FILENAME = "bootstrap.properties";
 
     private static Set<String> customServerEnvFiles = new HashSet<String>();
     private static Set<String> customBootstrapFiles = new HashSet<String>();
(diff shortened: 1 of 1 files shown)


lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/codeactions/AddFeature.java
```diff
@@ -123,7 +123,14 @@ public void doCodeAction(ICodeActionRequest request, List<CodeAction> codeAction
             referenceRangeStart = server.getStart();
             referenceRangeEnd = server.getStartTagCloseOffset()+1;
         }
+        
         Range referenceRange = XMLPositionUtility.createRange(referenceRangeStart, referenceRangeEnd, document);
+        // check if null, clarify and reduce verbosity of error if featureManager not found.
+        // TODO: can check earlier to avoid createRange()'s error log above
+        if (referenceRange == null) {
+            LOGGER.warning("Could not find range for feature insertion. featureManager may not be defined.");
+            return;
```
cherylking, line 132: I don't understand this new code. We won't even come in here if there is no `featureManager` with the current implementation. Is this for the future?

liberty-ls/src/main/java/io/openliberty/tools/langserver/LibertyConfigFileManager.java
```diff
@@ -105,25 +105,17 @@ public static boolean isServerEnvFile(LibertyTextDocument tdi) {
...
-        if (filePath.endsWith(DEFAULT_SERVER_ENV) || customServerEnvFiles.contains(filePath)) {
-            return true;
-        }
-
-        File file = new File(filePath);
-        if (file.getName().equals("server.env")) {
-            return customConfigDirs.contains(file.getParent());
-        }
-        return false;
+        return filePath.endsWith(SERVER_ENV_FILENAME) ||
```
cherylking, line 117: It is better to check the file name for a direct match than `endsWith`, unless you include the path separator (which you need to be careful with on different platforms...do we only have forward slashes because this is a `uri`?).

cherylking (changes requested):

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/codeactions/AddFeature.java
evie-lau: Removing this to debug further and address separately. Only happens in a strange scenario with an improperly XML-formatted server config xml.

cherylking (approved):

PR #240, opened 2023-11-01
Add git.properties file with build id

scottkurz:


cherylking:
Is this going to play well with the `maven-shade-plugin` we are using to control what goes into the build artifact?

scottkurz:
> Is this going to play well with the `maven-shade-plugin` we are using to control what goes into the build artifact?

I think it does, for now...the git.properties file still seems to appear in the final artifact.

PR #243, opened 2023-11-03
Rework how Liberty config xml files are determined

### LINKED ISSUE #242: SVT:Hovering and feature completion not working in server.xml
I have Liberty Tools installed in VSCode in Mac, with my Acme application.

When I try to see the features in server.xml, hovering or feature completion are not working.

image

cherylking:
Fixes #242

### DIFF
--- a/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/XmlReader.java
+++ b/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/XmlReader.java
@@ -51,31 +51,42 @@ public class XmlReader {
     }
 
     public static boolean hasServerRoot(File xmlFile) {
-        XMLEventReader reader = null;
+        if (!xmlFile.exists() || xmlFile.length() == 0) {
+            return false;
+        }
         
         try {
-            if (!xmlFile.exists() || xmlFile.length() == 0) {
-                return false;
+            XMLInputFactory factory = XMLInputFactory.newInstance();
+            try {
+                factory.setProperty(XMLInputFactory.IS_VALIDATING, Boolean.FALSE);
+                factory.setProperty(XMLInputFactory.IS_SUPPORTING_EXTERNAL_ENTITIES, Boolean.FALSE);
+                factory.setProperty(XMLInputFactory.SUPPORT_DTD, Boolean.FALSE);
+            } catch (Exception e) {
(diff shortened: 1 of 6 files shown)


lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/XmlReader.java
```diff
@@ -51,31 +51,42 @@ public static boolean hasServerRoot(Path filePath) {
...
+
+            XMLEventReader reader = null;
+
+            try (FileInputStream fis = new FileInputStream(xmlFile)) {
+                reader = factory.createXMLEventReader(fis);
+                while (reader.hasNext()) {
+                    XMLEvent nextEvent = reader.nextEvent();
+                    if (nextEvent.isStartElement()) {
+                        return isServerElement(nextEvent);
+                    }
```
evie-lau, line 76: I know we never accounted for this before, but does anyone ever add an [XML prolog](https://www.w3schools.com/XML/xml_syntax.asp) to a server config xml?
Wondering if this would fail it, even if there is a `<server>` root tag after the XML prolog.

ie. 
```
<?xml version="1.0" encoding="UTF-8"?> 
<server>
...
</server>
```

evie-lau (changes requested):

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/XmlReader.java
evie-lau: I also noticed the switch from `nextTag()` to `nextEvent()`. It seems `nextTag()` will skip some insignificant events until a start/end element is found.

cherylking: It handles this because it only looks at the first start element, and that is not a start element. It is special. I can add one to an existing test case to prove it.

cherylking: Actually, we already have `src/test/resources/sample/custom_server.xml` that contains that XML prolog. It is used in `XmlReaderTest` and passes both the `hasServerRoot` check and `isConfigXMLFile` check.

evie-lau (approved):

PR #257, opened 2023-12-04
Avoid class variables in diagnostic code

### LINKED ISSUE #256: Incorrect diagnostics  "ERROR: cdi-4.0 is already included."
With LCLS v2.1.1 in LTE 23.0.12 we sometimes get diagnostics like:  
  "ERROR: cdi-4.0 is already included." with the red X and underline in server.xml

I watched this happen in the debugger.

The problem centers around the same LibertyDiagnosticParticipant instance getting used across two threads, with a field holding the features:  `private Set<String> includedFeatures;`.

This would seem to need to be refactored to use a local/stack variable.   

I'm taking for granted there's nothing wrong with the high-level flow re: why we're calling diagnostics across 2 threads in the first place.  I've no reason to think that's wrong, but let me know if anyone disagrees.

cherylking:
Fixes #256


scottkurz (approved):
LGTM

PR #283, opened 2024-08-06
adding property to xmlreader of lemminx

arunvenmany-ibm:


lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/DocumentUtil.java
```diff
@@ -87,6 +87,8 @@ public static void writeDocToXmlFile(Document doc, File inputFile) throws Except
     private static TransformerFactory getTransformerFactory() throws TransformerConfigurationException {
         TransformerFactory transformerFactory = TransformerFactory.newInstance();
         transformerFactory.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, Boolean.TRUE);
+        transformerFactory.setAttribute(XMLConstants.ACCESS_EXTERNAL_DTD, "");
```
cherylking, line 90: Setting these to empty string seems odd. Are we sure that is correct? Do you have some javadoc or example/sample that shows we should do that?

cherylking (commented):
two questions

PR #301, opened 2024-09-05
microprofile platform diagnostic issue fix

arunvenmany-ibm:
for microProfile, found that platforms were not getting matched. found the issue was related to lowercase conversion

Added test from https://openliberty.io/docs/latest/reference/feature/versionless-features.html

### DIFF
--- a/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
+++ b/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
@@ -344,7 +344,7 @@ public class LibertyDiagnosticParticipant implements IDiagnosticsParticipant {
                     list.add(new Diagnostic(range, message, DiagnosticSeverity.Error, LIBERTY_LEMMINX_SOURCE));
                 }
                 preferredPlatformsWithoutVersion.add(platformNoVersionLower);
-                preferredPlatforms.add(platformNameLowerCase);
+                preferredPlatforms.add(platformName);
 
             }
         }
--- a/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
+++ b/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
@@ -344,7 +344,7 @@ public class LibertyDiagnosticParticipant implements IDiagnosticsParticipant {
                     list.add(new Diagnostic(range, message, DiagnosticSeverity.Error, LIBERTY_LEMMINX_SOURCE));
                 }
                 preferredPlatformsWithoutVersion.add(platformNoVersionLower);
-                preferredPlatforms.add(platformName);
+                preferredPlatforms.add(platformNameLowerCase);
 
             }
         }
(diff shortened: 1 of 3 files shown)


lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
```diff
@@ -344,7 +344,7 @@ private void validatePlatform(DOMDocument domDocument, List<Diagnostic> list, DO
                     list.add(new Diagnostic(range, message, DiagnosticSeverity.Error, LIBERTY_LEMMINX_SOURCE));
                 }
                 preferredPlatformsWithoutVersion.add(platformNoVersionLower);
-                preferredPlatforms.add(platformNameLowerCase);
+                preferredPlatforms.add(platformName);
```
cherylking, line 347: There's a reason we store them in lowercase. Liberty let's the user enter feature names and I presume platform names in any case they want. We should do all comparisons/matching based on lower case names. It's the only way we can ensure proper functionality.

cherylking (changes requested):

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
arunvenmany-ibm: got it.. changed in the logic where we fetch the platforms from json. changed to lowercase

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/FeatureService.java
```diff
@@ -634,7 +634,7 @@ public Set<String> getAllPlatformsForVersionLessFeature(String featureName, Stri
         }
         return featureNames.stream()
                 .map(f -> getAllPlatformsForFeature(f, libertyVersion, libertyRuntime, requestDelay, documentURI))
-                .flatMap(Set::stream).collect(Collectors.toSet());
+                .flatMap(Set::stream).map(String::toLowerCase).collect(Collectors.toSet());
```
cherylking, line 637: so the fix isn't this easy because we should preserve case (for doing completion) and ignore case for doing matching/validation

cherylking (changes requested):

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/FeatureService.java
arunvenmany-ibm: yes this method is only for matching and validation. This method is used to fetch all platforms for a versionless feature
For platform completion and platform diagnostic, we use different method

cherylking: interesting...I would not have thought we need different methods (except for the upper case vs lower case nuance...which could be handled by a boolean flag parameter). Which method is used for completion?

cherylking: I see this is specific to platforms for a specific versionless feature.

cherylking (approved):

PR #305, opened 2024-09-30
fixing feature uniqueness diagnostic issue for renamed features

arunvenmany-ibm:


lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
```diff
@@ -163,14 +167,31 @@ private void validateFeature(DOMDocument domDocument, List<Diagnostic> list, Set
...
+                        entry -> entry.getValue().toLowerCase()));
+        Map<String, String> changedFeatureNameMapLowerReversed =
+                changedFeatureNameMapLower.entrySet()
+                        .stream()
+                        .collect(Collectors.toMap(Map.Entry::getValue, Map.Entry::getKey));
+        Set<String> featuresWithOldNames = featuresWithoutVersions.stream().map(
+                        v -> changedFeatureNameMapLower.get(v + "-"))
+                .filter(Objects::nonNull)
+                .collect(Collectors.toSet());
+        Set<String> featuresWithChangedNames = featuresWithoutVersions.stream().map(
```
cherylking, line 191: I don't think we really need to check both directions for this. If when we find a feature with the old name, we check for existence of any features with the new name and flag the diagnostic on the old name, that should be sufficient. That way we don't have to do the reverse lookup.

lemminx-liberty/src/test/java/io/openliberty/LibertyDiagnosticTest.java
```diff
@@ -687,4 +687,51 @@ public void testConfigElementVersionLess() throws JAXBException {
...
+                "<server description=\"Sample Liberty server\">", //
+                "       <featureManager>", //
+                "               <feature>jms-2.0</feature>", //
+                "               <feature>messaging-3.0</feature>", //
+                "       </featureManager>", //
+                "</server>" //
+        );
+        Diagnostic invalid = new Diagnostic();
+        invalid.setRange(r(3, 24, 3, 37));
+        invalid.setMessage("ERROR: The messaging-3.0 feature cannot be configured with the jms-2.0 feature because they are two different versions of the same feature. The feature name changed from jms to messaging for Jakarta EE. Remove one of the features.");
```
cherylking, line 703: Can you use a feature with mixed case to prove that the code issuing this diagnostic is preserving the case of the feature names?

cherylking (changes requested):

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
arunvenmany-ibm: The current way of feature validation is linear and each feature will be added to included features only after each iteration. Thus we dont have a map which contains all features in the document

If we change the logic to only one way lookup,

For scenario 1, ie, new name is added first and then a old feature is coming, er the logic you suggested, this will show diagnostic 
For scenario 2, old feature is added first and then when a new name appears , per the logic you suggested, this wont show diagnostic, because we dont have a list of all features in document(we only have a list of features which is already validated)

lemminx-liberty/src/test/java/io/openliberty/LibertyDiagnosticTest.java
arunvenmany-ibm: Already added a test for this with ejb and enterpriseBeans . But this was versionless. I will add one more with specifying version for ejb and enterpriseBeans

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
cherylking: Got it...so it depends on which feature is found first. Does that mean we will most likely only get one message for this, but which feature gets flagged depends on their order in the list?

cherylking: btw...we could do this check after the list of features is collected at the end of the method `validateFeaturesAndPlatforms`...but it would require iterating through the includedFeatures which probably wouldn't be worth it since you can calculate the diagnostic the way you are currently.

lemminx-liberty/src/test/java/io/openliberty/LibertyDiagnosticTest.java
```diff
@@ -718,6 +718,22 @@ public void testUniquenessForNameChangedFeatureDiagnostic() throws BadLocationEx
...
+                "               <platform>javaee-7.0</platform>", //
+                "               <feature>enterpriseBeans-4.0</feature>", //
+                "               <feature>ejb-3.2</feature>", //
+                "       </featureManager>", //
+
+                "</server>" //
+        );
+        invalid = new Diagnostic();
+        invalid.setRange(r(4, 24, 4, 31));
+        invalid.setMessage("ERROR: The ejb-3.2 feature cannot be configured with the enterprisebeans-4.0 feature because they are two different versions of the same feature. The feature name changed from ejb to enterprisebeans for Jakarta EE. Remove one of the features.");
```
cherylking, line 733: I'm glad I had you add this test. The feature name `enterprisebeans-4.0` should be `enterpriseBeans-4.0`. It is not ok for our message to have the wrong case for the feature. This means your code needs to change to be able to use the specified feature name, not the lowercase version of it.

cherylking (changes requested):

lemminx-liberty/src/test/java/io/openliberty/LibertyDiagnosticTest.java
arunvenmany-ibm: Done

cherylking (approved):

PR #307, opened 2024-10-08
junit 4 to junit 5 changes for liberty-ls

arunvenmany-ibm:


cherylking:
I actually don't want to merge this in before our next release, because it introduces new OSS dependency. Need to get clearance and update NOTICE file etc.

PR #310, opened 2024-10-11
Server env blank line issue fix

arunvenmany-ibm:


cherylking:
Fix for https://github.com/OpenLiberty/liberty-tools-intellij/issues/1017

liberty-ls/src/test/java/io/openliberty/tools/langserver/diagnostic/ServerEnvDiagnosticTest.java
```diff
@@ -27,7 +27,7 @@ public void testServerEnv() throws Exception {
             // Checking invalid whitespace before equal sign: WLP_DEBUG_REMOTE =n
             createRange(4,16,18),
             // Checking invalid whitespace after equal sign: WLP_LOGGING_MESSAGE_FORMAT= SIMPLE
-            createRange(5,26,28)  
+            createRange(7,26,28)
```
cherylking, line 30: Don't replace the test for line 5 with line 7, just add a new one. Also, it seems like line 8 was not tested. The numbers are off by one I think....if line 1 is actually `0` for `createRange`, then we should be testing the following lines: 0, 2, 3, 4, 5, 6 and 7
And add appropriate comment before each `createRange` on what it is testing.

cherylking (changes requested):

liberty-ls/src/test/java/io/openliberty/tools/langserver/diagnostic/ServerEnvDiagnosticTest.java
cherylking: Never mind. I see now that there should be no diagnostic for lines 5 (comment line) and 6 (blank line) and what was tested at line 5 is now line 7.

cherylking (approved):

PR #325, opened 2024-12-09
Var processing 2.0

arunvenmany-ibm:


lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyCompletionParticipant.java
```diff
@@ -53,12 +53,24 @@ public void onAttributeValue(String valuePrefix, ICompletionRequest request, ICo
...
                         .getDocumentURI());
-        String variableName = valuePrefix.replace("$", "")
-                .replace("{", "")
-                .replace("}", "");
-        variableProps.entrySet().stream().filter(it -> it.getKey().toString().toLowerCase().contains(variableName.toLowerCase()))
+        // value prefix will contain all existing attribute value, till the cursor
+        String variableName = valuePrefix;
+        String replacePrefix;
+        if (valuePrefix.contains("${")) {
+            variableName = valuePrefix.substring(valuePrefix.lastIndexOf("${"))
```
cherylking, line 60: I feel like this won't work if the user entered something like `${myFirstVar}/mySecond` and triggered completion at the end of the String. This code will decide to do completion on `myFirstVar` won't it? When you demoed it, I thought it worked right, but from looking at this code I don't think so.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyCompletionParticipant.java
```diff
@@ -67,7 +79,7 @@ public void onAttributeValue(String valuePrefix, ICompletionRequest request, ICo
                     completionItem.setTextEdit(edit);
                     completionItem.setFilterText(variableProp.getKey().toString());
                     completionItem.setKind(CompletionItemKind.Value);
-                    completionItem.setDocumentation(String.format("%s = %s", variableProp.getKey(),variableProp.getValue()));
+                    completionItem.setDocumentation(String.format("%s = %s", variableProp.getKey(), variableProp.getValue()));
```
cherylking, line 82: Is this showing the value during completion? I thought we decided against that.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyExtension.java
```diff
@@ -77,6 +80,19 @@ public void start(InitializeParams initializeParams, XMLExtensionsRegistry xmlEx
         } catch (Exception e) {
             throw new RuntimeException(e);
         }
+
+        // for each workspace, a file alteration observer is added
+        for (LibertyWorkspace workspace : LibertyProjectsManager.getInstance().getLibertyWorkspaceFolders()) {
+            // checking for any changes in target folder
+            Path libertyPath = new File(workspace.getWorkspaceURI().getPath(),
+                    "target").toPath();
```
cherylking, line 88: We have to support Gradle as well. For Gradle projects, we would use `build` folder. I am worried about monitoring the entire target/build folder. I feel like we should be able to be more selective. We only care about the Liberty server folder. In the case that the `liberty-plugin-config.xml` already exists, we can figure out which folder we actually need to monitor (the `serverDirectory`).

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyConstants.java
```diff
@@ -74,4 +76,6 @@ private LibertyConstants() {
 
     public static String changedFeatureNameDiagMessage="ERROR: The %s feature cannot be configured with the %s feature because they are two different versions of the same feature. " +
             "The feature name changed from %s to %s for Jakarta EE. Remove one of the features.";
+
+    public static List<String> filesToWatch= Arrays.asList(SERVER_XML,"server.env","bootstrap.properties");
```
cherylking, line 80: We actually care about all xml files since the server.xml can include other xml files. There are several folders that can contain xml files used for server configuration. See this [doc](https://openliberty.io/docs/latest/reference/config/server-configuration-overview.html#server-xml).

cherylking (changes requested):
I have not reviewed the tests yet.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyCompletionParticipant.java
arunvenmany-ibm: for showing completion of second variable, variable should always prefix with ${ . Hence show no completion here

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyCompletionParticipant.java
arunvenmany-ibm: It is showing only in documentation section of completion, not in actual value
image

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyConstants.java
arunvenmany-ibm: We are watching other config files separately using LibertyUtils.isConfigXMLFile, which include pattern
".*/(?:usr/shared/config/.+|configDropins/(?:defaults|overrides)/.+|server\\.xml)$";

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyExtension.java
arunvenmany-ibm: We were monitoring the target folder because we cannot find serverDirectory without creating liberty-plugin-config.xml file

Watch file is added entirely to support file creation/updating/deletion for config file. But if dev mode is not started, we wont know the serverDirectory so we don't know which path to look for .
Only thing constant in the directory path would be "build" or "target" folder and rest will differ based on serverName.

I will add build folder as well into the watch list

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyCompletionParticipant.java
```diff
@@ -60,7 +60,7 @@ public void onAttributeValue(String valuePrefix, ICompletionRequest request, ICo
             variableName = valuePrefix.substring(valuePrefix.lastIndexOf("${"))
                     .replace("${", "");
             if (variableName.contains("}")) {
-                variableName = variableName.replace("}", "");
+                variableName = variableName.substring(variableName.lastIndexOf("}") + 1);
```
cherylking, line 63: if nothing appears after the last `}`, then we want to do completion on what was in the curly braces don't we?

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/FileWatchService.java
```diff
@@ -68,7 +68,10 @@ public void addFileAlterationObserver(LibertyWorkspace workspace, List<String> w
 
     private FileAlterationObserver getFileAlterationObserver(final String parentPath, LibertyWorkspace workspace) {
         IOFileFilter notFileFilter = FileFilterUtils.notFileFilter(
-                new SuffixFileFilter(Arrays.asList(".class", ".lst", ".txt"), IOCase.INSENSITIVE));
+                new SuffixFileFilter(Arrays.asList(".class", ".lst", ".txt",".log",".manager",".libertyls",".sLock"),
```
cherylking, line 71: I like this!

I think we can add some more file extensions here: .jar, .war, .ear

Also some dirs: libs, tmp, classes, generated-sources, generated-test-sources, invoker-reports, it, maven-status, surefire-reports, test-classes

Do you see a down side to adding these?

cherylking (changes requested):

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyCompletionParticipant.java
```diff
@@ -53,12 +54,33 @@ public void onAttributeValue(String valuePrefix, ICompletionRequest request, ICo
...
+        //getting all existing variables in current completion prefix string
+        List<VariableLoc> variables = LibertyUtils.getVariablesFromTextContent(valuePrefix);
+        // value prefix will contain all existing attribute value, till the cursor
+        String variableName = valuePrefix.replace("${", "")
                 .replace("}", "");
-        variableProps.entrySet().stream().filter(it -> it.getKey().toString().toLowerCase().contains(variableName.toLowerCase()))
+        StringBuilder prefixVariables = new StringBuilder();
+        String replacePrefix = "";
+        for (VariableLoc variableLoc : variables) {
+            prefixVariables.append(String.format("${%s}", variableLoc.getValue()));
```
cherylking, line 65: I don't understand this prefixVariables logic. It seems like you are iterating through all the variables found in the passed in valuePrefix and arbitrarily appending them to each other without regard to other characters that may have occurred between variable references in the original valuePrefix.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyCompletionParticipant.java
```diff
@@ -53,12 +54,33 @@ public void onAttributeValue(String valuePrefix, ICompletionRequest request, ICo
         Properties variableProps = SettingsService.getInstance()
                 .getVariablesForServerXml(request.getXMLDocument()
                         .getDocumentURI());
-        String variableName = valuePrefix.replace("$", "")
-                .replace("{", "")
+        //getting all existing variables in current completion prefix string
+        List<VariableLoc> variables = LibertyUtils.getVariablesFromTextContent(valuePrefix);
```
cherylking, line 58: I went and looked at what this method is returning and noticed a couple of things that I should have commented previously. The two constants for the regex and pattern should have names that reflect what they are and should be all capital. See the `EXCLUDE_PATTERN_REGEX` and `EXCLUDE_PATTERN` constants in LibertyUtils. Change the names of `regex` and `p` accordingly. 

Now a quick question....is the getVariablesFromTextContent method returning just what was inside `${...}` without the beginning and ending syntax?

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyCompletionParticipant.java
```diff
@@ -53,12 +54,33 @@ public void onAttributeValue(String valuePrefix, ICompletionRequest request, ICo
...
+        String variableName = valuePrefix.replace("${", "")
                 .replace("}", "");
-        variableProps.entrySet().stream().filter(it -> it.getKey().toString().toLowerCase().contains(variableName.toLowerCase()))
+        StringBuilder prefixVariables = new StringBuilder();
+        String replacePrefix = "";
+        for (VariableLoc variableLoc : variables) {
+            prefixVariables.append(String.format("${%s}", variableLoc.getValue()));
+        }
+        if (!variables.isEmpty()) {
+            // find last variable last character and consider new variable completion from that location
```
cherylking, line 68: Do we know the actual position of the cursor when they triggered completion?

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
```diff
@@ -129,7 +129,7 @@ private void validateVariables(DOMDocument domDocument, List<Diagnostic> diagnos
         for (VariableLoc variable : variables) {
             if (!variablesMap.containsKey(variable.getValue())) {
                 String variableInDoc = String.format("${%s}", variable.getValue());
-                Range range = XMLPositionUtility.createRange(variable.getStartLoc(),variable.getEndLoc(),
+                Range range = XMLPositionUtility.createRange(variable.getStartLoc()-2,variable.getEndLoc()+1,
```
cherylking, line 132: Add comment to the code on why you are doing `-2` and `+1` here. Give a concrete example.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyHoverParticipant.java
```diff
@@ -59,7 +59,7 @@ public Hover onAttributeValue(IHoverRequest request, CancelChecker cancelChecker
                 stringBuilder.append(String.format("%s = %s", variable.getValue(), variableMap.get(variable.getValue())));
             }
             if (varIter.hasNext()) {
-                stringBuilder.append(System.lineSeparator());
+                stringBuilder.append("<br />");
```
cherylking, line 62: Is the blank space in the string intentional?

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/codeactions/ReplaceVariable.java
```diff
@@ -64,10 +66,12 @@ public void doCodeAction(ICodeActionRequest request, List<CodeAction> codeAction
...
                 }
-                /*for (Map.Entry<Object, Object> nextVariable : existingVariables.entrySet()) {
-                    String title = "Replace Variable with " + nextVariable.getKey() + " with value = " + nextVariable.getValue();
-                    codeActions.add(CodeActionFactory.replace(title, diagnostic.getRange(), nextVariable.getKey().toString(), document.getTextDocument(), diagnostic));
-                }*/
+                // use special code action to reload variable diagnostics
+                Range range = XMLPositionUtility.createRange(document.getDocumentElement().getEndTagCloseOffset(), document.getDocumentElement().getEndTagCloseOffset()+1,
+                        document);
+                String title = "Reload Diagnostics";
+                codeActions.add(CodeActionFactory.insert(title, range.getEnd(), " ",
```
cherylking, line 73: I don't think we should do this as we discussed in our last scrum.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/FileWatchService.java
```diff
@@ -0,0 +1,149 @@
...
+        }
+    }
+
+    private FileAlterationObserver getFileAlterationObserver(final String parentPath, LibertyWorkspace workspace) {
+        //ignore files/directories with below suffixes and names
+        IOFileFilter notFileFilter = FileFilterUtils.notFileFilter(
+                new SuffixFileFilter(Arrays.asList(".class", ".lst", ".txt", ".log", ".manager", ".libertyls",
+                        ".sLock", ".jar", ".war", ".ear",".mf"),
+                        IOCase.INSENSITIVE)
+                        .or(new NameFileFilter(Arrays.asList("workarea", "plugin-cfg.xml", "libs", "tmp", "classes","bin",
```
cherylking, line 75: Is `plugin-cfg.xml` really a directory? Seems like you changed this logic...are these just files now? Before you added files in one part and directories in another. Is there no distinction?

cherylking (changes requested):

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/FileWatchService.java
arunvenmany-ibm: NameFileFilter accepts both files and directories
And it looks for base names from the path, but only considers the path after our watch location. In below example, we watch target folder and I have set path for my workspace as  "/Users/arunkumarvn/Downloads/libs/app-name/target/liberty/wlp"

when an item present in our namefilter is found, ie generated-sources, accept() returns true
image
image
image


whereas if the current file is not part of any dir/file specified in namefilter, accept() returns false
image

image

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyCompletionParticipant.java
arunvenmany-ibm: I have modified the logic to support 4 scenarios

1. User already has multiple variables, added a new variable for completion
  a. User entered completion data  ${testVar1}/abc| -> in this case /abc is considered as completion variable and 
 probably there is no variable which contains /abc , hence no completion will be shown
  b. User entered completion data  ${testVar1}/${abc| -> in this case we identify the second variable as abc, tries to find any completion data which contains abc, and it will be shown with prefix ${testVar1}/
  c. User entered completion data  ${testVar1}/apps/${testVar2}/${abc|  -> same as case b, completion will be shown with prefix ${testVar1}/apps/${testVar2}/
  
  2. User do not have any existing variable in current attribute value
   a. User entered completion data "/apps" -> completion will be shown if there is any variable which contains "/apps"
   b. User entered completion data "/apps/${a" -> we identify ${ as starting of a variable, extract "a" as variable completion item, completes data based on any variable which contains prefix "/apps"

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyCompletionParticipant.java
arunvenmany-ibm: Corrected variable names.
getVariablesFromTextContent returns variable value inside ${} and start and end index

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyCompletionParticipant.java
arunvenmany-ibm: In the onAttributeValue() method, first argument is the prefix variable. ie, if user has entered ${abc|, prefix will contain ${abc

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
arunvenmany-ibm: added
//range is used in ReplaceVariable to provide quick fix.
// we just need the variable value range here as ${} is added in replace variable message

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyHoverParticipant.java
arunvenmany-ibm: It was intentional . <br /> is the XHTML compatible way. Normal System.lineSeparator() was not working in our client side hence I tried with <br /> and it works

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/FileWatchService.java
cherylking: That's fine. We don't care about anything in these subfolders.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyCompletionParticipant.java
```diff
@@ -56,32 +56,42 @@ public void onAttributeValue(String valuePrefix, ICompletionRequest request, ICo
...
-                // add char between last variable and start of completion variable to prefix
-                replacePrefix = valuePrefix.substring(lastVar.getEndLoc() + 1, valuePrefix.lastIndexOf("${"));
+            // if last index of an existing variable is less than lastIndex of ${,
+            // this means that completion variable is started with ${
+            if (valuePrefix.lastIndexOf("${") > lastVar.getEndLoc()) {
+                // get whatever is after last ${ as variablePrefix for Completion
+                variablePrefix = valuePrefix.substring(valuePrefix.lastIndexOf("${") + 2);
+                completionPrefix = valuePrefix.substring(0, valuePrefix.lastIndexOf("${"));
+            } else {
+                variablePrefix = valuePrefix.substring(lastVar.getEndLoc() + 1);
```
cherylking, line 73: Does this handle the case where the user types `${}` and then goes inside the brackets and adds characters and triggers completion? So when they trigger completion it is `${someChars|}`. In this case, there will be one variable in the `variables` collection and the `lastIndexOf("${")` will not be > than the `lastVar.getEndLoc()`. Then this logic at line 77 will set the `variablePrefix` to the closing `}`?

cherylking (changes requested):

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyCompletionParticipant.java
cherylking: Do we have a test case for `${someVar|}`?

arunvenmany-ibm: Yes
image

cherylking (approved):

PR #332, opened 2025-01-02
fixing completion issues for features when no input is provided

### LINKED ISSUE #275: Server.xml completion can include features with incompatible version numbers
If I start typing a feature name and use completion the server will not suggest features which are not compatible with existing features. In this case we do not see other versions of `jakartaee`.
image
However if I do not start typing a feature name the server will give the full list of available features even ones incompatible with existing features. I would expect similar filtering to the previous example.
image
The problem becomes worse when I type `ja` because local filtering then displays the incompatible features.
image

arunvenmany-ibm:
Fixes #275


arunvenmany-ibm:
attaching test video
https://github.com/user-attachments/assets/aed34185-7718-47b5-ae9e-0fe4bdcf86a2

I have added AppAuthentication-3.0 and then tried to find completion in 2 scenarios
1. by entering 'app' -> appAuthentication versions are not showing up
2. by entering '' -> appAuthentication versions are not showing up

PR #338, opened 2025-01-28
 quickfix code actions for invalid property values

### LINKED ISSUE #313: Diagnostic messages does't show any valid values when an invalid value is given for a variable in server.env
When an invalid value is given for a property the following diagnostic message is shown 
image
Here it does't mention any valid values 

however on hover over the property it displays the valid values as below 
image

it will be more helpful to show the valid values along with the diagnostic message if an invalid input is given by the user , he might not be aware about the hover message over the property. 

Alternatively if quick fixes are available along  with the diagnostic messages then it can also be a possible workaround

arunvenmany-ibm:
Fixes #313


arunvenmany-ibm:
QuickFix Code action demo from intellij
https://github.com/user-attachments/assets/0c798684-8f7d-4b5d-a0a4-45daf2a6ea47

QuickFix Code action demo from vscode

https://github.com/user-attachments/assets/5fa875c8-eca1-4dbf-89fb-11b51d69f3d0

PR #340, opened 2025-01-28
fixing platform quickfix code completion not working

### LINKED ISSUE #334: Make platform element perform similar to feature element for diagnostic/quick fixes
For a feature element, a quick fix is provided when the feature version does not match an known feature. 

Screenshot 2025-01-13 at 8 19 39 AM

For the platform element, there is just a diagnostic and no quick fix.

Screenshot 2025-01-13 at 8 20 00 AM

I think we should provide a quick fix for the platform element in this case, showing all the possible choices for that platform. Also, it should not show a description for the platform element that is not a valid platform.

arunvenmany-ibm:
Fixes #334


arunvenmany-ibm:
Now platform quick fix is showing up 

https://github.com/user-attachments/assets/e0b904ce-b1c3-44a8-8ada-f75667b6bf27

cherylking (approved):

PR #341, opened 2025-01-29
changes to resolve config element problem

### LINKED ISSUE #335: Incorrect config element diagnostic with versionless feature
This `mpMetrics` config element should not have a diagnostic since the versionless `mpMetrics` feature is specified.

Screenshot 2025-01-13 at 10 31 28 AM

arunvenmany-ibm:
Fixes #335

### DIFF
--- a/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/data/FeatureListGraph.java
+++ b/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/data/FeatureListGraph.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright (c) 2023, 2024 IBM Corporation and others.
+* Copyright (c) 2023, 2025 IBM Corporation and others.
 *
 * This program and the accompanying materials are made available under the
 * terms of the Eclipse Public License v. 2.0 which is available at
@@ -55,7 +55,7 @@ public class FeatureListGraph {
         }
         FeatureListNode node = new FeatureListNode(nodeName, description);
         featureNodes.put(nodeName, node);
-        nodes.put(nodeName, node);
+        nodes.putIfAbsent(nodeName, node);
         return node;
     }
     
@@ -65,7 +65,7 @@ public class FeatureListGraph {
         }
         ConfigElementNode node = new ConfigElementNode(nodeName);
         configElementNodes.put(nodeName, node);
-        nodes.put(nodeName, node);
+        nodes.putIfAbsent(nodeName, node);
         return node;
     }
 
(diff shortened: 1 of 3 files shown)


lemminx-liberty/src/test/java/io/openliberty/LibertyDiagnosticTest.java
```diff
@@ -1062,4 +1062,74 @@ public void testMultipleVariablesInSameAttributeDiagnostic() {
...
+        List<String> featuresToAdd = new ArrayList<String>();
+        featuresToAdd.add("mpMetrics-1.0");
+        featuresToAdd.add("mpMetrics-1.1");
+        featuresToAdd.add("mpMetrics-2.0");
+        featuresToAdd.add("mpMetrics-2.2");
+        featuresToAdd.add("mpMetrics-2.3");
+        featuresToAdd.add("mpMetrics-3.0");
+        featuresToAdd.add("mpMetrics-4.0");
+        featuresToAdd.add("mpMetrics-5.0");
+        featuresToAdd.add("mpMetrics-5.1");
```
cherylking, line 1111: So this is interesting....we are suggesting they add a versioned feature...but they have a platform specified already. Which means it would be better to add the versionless feature. We probably should open an enhancement issue to look into making our quick fix smarter for this scenario.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/data/FeatureListGraph.java
```diff
@@ -65,7 +70,7 @@ public ConfigElementNode addConfigElement(String nodeName) {
         }
         ConfigElementNode node = new ConfigElementNode(nodeName);
         configElementNodes.put(nodeName, node);
-        nodes.putIfAbsent(nodeName, node);
+        nodes.put(nodeName, node);
```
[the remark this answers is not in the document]
arunvenmany-ibm: Q1 :- Yes correct
Q2:- Yes I think its okay to go ahead with this, as we are not expecting versionless to have separate enable By options
Q3:- We can modify the logic to ignore versionless feature. But again, this issue is only for mpMetrics , that is the only case where configElementName and featureName are same

lemminx-liberty/src/test/java/io/openliberty/LibertyDiagnosticTest.java
arunvenmany-ibm: Done
https://github.com/OpenLiberty/liberty-language-server/issues/342

cherylking (approved):

PR #343, opened 2025-02-03
Multi value property fix

### LINKED ISSUE #336: server.env validation reports error on multiple console source
The liberty language server indicates some errors on 
`WLP_LOGGING_CONSOLE_SOURCE=message,trace,accessLog,ffdc,audit`
see 

Image

although it's a legal setting documented here:
https://openliberty.io/docs/latest/log-trace-configuration.html#container

Liberty plugin : 24.0.12

arunvenmany-ibm:
fixes #336

### DIFF
--- a/liberty-ls/src/test/java/io/openliberty/tools/langserver/completion/BootstrapPropertyCompletionTest.java
+++ b/liberty-ls/src/test/java/io/openliberty/tools/langserver/completion/BootstrapPropertyCompletionTest.java
@@ -141,6 +141,10 @@ public class BootstrapPropertyCompletionTest extends AbstractCompletionTest {
 
         completions = getCompletion("com.ibm.ws.logging.console.source=trace,message", new Position(0, 41));
         completionItems = completions.get().getLeft();
+        assertEquals(1, completionItems.size());
+
+        completions = getCompletion("com.ibm.ws.logging.message.source=trace,messag,", new Position(0, 40));
+        completionItems = completions.get().getLeft();
         assertEquals(0, completionItems.size());
     }
 
--- a/liberty-ls/src/test/java/io/openliberty/tools/langserver/completion/BootstrapPropertyCompletionTest.java
+++ b/liberty-ls/src/test/java/io/openliberty/tools/langserver/completion/BootstrapPropertyCompletionTest.java
@@ -143,6 +143,7 @@ public class BootstrapPropertyCompletionTest extends AbstractCompletionTest {
         completionItems = completions.get().getLeft();
         assertEquals(1, completionItems.size());
 
+        // return no completion because messag is invalid
         completions = getCompletion("com.ibm.ws.logging.message.source=trace,messag,", new Position(0, 40));
         completionItems = completions.get().getLeft();
         assertEquals(0, completionItems.size());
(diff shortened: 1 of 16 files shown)


arunvenmany-ibm:
writing down newly developed behaviors below for multi values variables

1. Completion

- When user trigger completion with single value, behavior is similar to any other single value field
image
- When user enters a comma after first value and do a completion, all the allowed
image
- When user add more values and trigger completion, result will show all available completions, excluding already selected
image
- If user enters an invalid value anywhere in the string and try completion, no completion will be show
image
image

2. Diagnostics
3. Code Action(QuickFix)

liberty-ls/src/test/java/io/openliberty/tools/langserver/completion/BootstrapPropertyCompletionTest.java
```diff
@@ -117,6 +117,37 @@ public void testValueCompletionLogProviderUserEnteredInvalidValue() throws Excep
...
+
+        completions = getCompletion("com.ibm.ws.logging.console.source=trace,", new Position(0, 34));
+        completionItems = completions.get().getLeft();
+        assertEquals(4, completionItems.size());
+        checkCompletionsContainAllStrings(completionItems, "trace,accessLog","trace,message","trace,ffdc","trace,audit");
+
+        completions = getCompletion("com.ibm.ws.logging.console.source=trace,a", new Position(0, 35));
+        completionItems = completions.get().getLeft();
+        assertEquals(3, completionItems.size());
+        checkCompletionsContainAllStrings(completionItems, "trace,accessLog","trace,audit","trace,message");
```
cherylking, line 135: Why would we return `trace,message` when the entered value was `trace,a`. Seems like we would only return the two where the second value started with `a`.

liberty-ls/src/test/java/io/openliberty/tools/langserver/completion/BootstrapPropertyCompletionTest.java
```diff
@@ -117,6 +117,37 @@ public void testValueCompletionLogProviderUserEnteredInvalidValue() throws Excep
...
+        completions = getCompletion("com.ibm.ws.logging.console.source=trace,message,", new Position(0, 42));
+        completionItems = completions.get().getLeft();
+        assertEquals(3, completionItems.size());
+        checkCompletionsContainAllStrings(completionItems, "trace,message,accessLog","trace,message,ffdc","trace,message,audit");
+
+        completions = getCompletion("com.ibm.ws.logging.console.source=trace,message", new Position(0, 41));
+        completionItems = completions.get().getLeft();
+        assertEquals(1, completionItems.size());
+
+        completions = getCompletion("com.ibm.ws.logging.message.source=trace,messag,", new Position(0, 40));
```
cherylking, line 146: Add a comment on why this returns no completions (because it has an invalid value of `messag` in it)..

cherylking (changes requested):
Love all the new method documentation.

liberty-ls/src/test/java/io/openliberty/tools/langserver/completion/BootstrapPropertyCompletionTest.java
arunvenmany-ibm: because completion is using 'contains'

screenshots f
https://github.com/OpenLiberty/liberty-language-server/blob/main/liberty-ls/src/main/java/io/openliberty/tools/langserver/model/propertiesfile/PropertiesKeyInstance.java


image

cherylking (approved):

PR #349, opened 2025-03-03
 changes to integrate liberty version 25.0.0.2

arunvenmany-ibm:
integrate liberty version 25.0.0.2 

LibertyFeatureTest updated to version 25.0.0.2 from 24.0.0.11, hence has more features enabled including mpopenapi-4.0, mpfaulttolerance-4.1, microprofile-7.0, mprestclient-4.0


lemminx-liberty/src/it/variable-processing-ol-it/pom.xml
```diff
@@ -84,7 +84,7 @@
                     <runtimeArtifact>
                         <groupId>io.openliberty</groupId>
                         <artifactId>openliberty-runtime</artifactId>
-                        <version>24.0.0.11</version>
+                        <version>25.0.0.2</version>
```
cherylking, line 87: For future reference, we do not need to update this version when updating cached data. It's important to prove we can generate for older runtimes and everything still works.

cherylking (approved):

PR #351, opened 2025-03-04
fixing diagnostics disappearing after hover

### LINKED ISSUE #350: Diagnostic is disappearing after hover
For variable resolution, if user have used variables in server.xml , but dev mode is not started, a diagnostic will appear on <server> element 

Image

But this disappears if we hover on some variable and do diagnostics again, even though dev mode is not started yet

Image

We depend upon dev mode to copy server.xml to liberty server directory and re populate variables.
We need a check before adding any variable to local map, so that without starting dev mode, new variables cannot be added

arunvenmany-ibm:
Fixes #350 

https://github.com/user-attachments/assets/807a9ee4-7293-46e1-ae5c-9b40ff79f9f5

### DIFF
--- a/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/SettingsService.java
+++ b/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/SettingsService.java
@@ -51,7 +51,7 @@ public class SettingsService {
 
     private Map<String,Properties> variables;
 
-    private boolean devModeRunning = false;
+    private boolean configCopiedToServer = false;
     /**
      * Takes the xml settings object and parses out the Liberty Settings
      * @param xmlSettings - All xml settings provided by the client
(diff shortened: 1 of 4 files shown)


lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/SettingsService.java
```diff
@@ -50,6 +50,8 @@ private SettingsService() {
     private LibertySettings settings;
 
     private Map<String,Properties> variables;
+
+    private boolean devModeRunning = false;
```
cherylking, line 54: There is nothing about the setting of this variable that truly indicates that dev mode is running (actively running). It seems you are taking the return of a non-empty Properties from the ci.common method to indicate that dev mode is running. Doesn't that only imply that some goal was run at some point on the project that caused the Liberty server to get created and the config copied over?

I think at best you could call this `libertyServerCreated` or `configCopiedToServer`.

cherylking (changes requested):

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/SettingsService.java
arunvenmany-ibm: Updated

cherylking (approved):

PR #355, opened 2025-03-13
changes to make sure variableMap is never null

### LINKED ISSUE #353: NullPointerException in LibertyDiagnosticParticipant.validateVariables method
See this NullPointerException stack trace which appears multiple times in the automated tests that run when building lemminx-liberty.

Image

The `variablesMap` should never be null. The method `SettingsService.getVariablesForServerXml` can return an empty Properties. At line 140, perhaps the Properties returned for the key are null? We should prevent that from happening, or return an empty Properties in that case. I worked on this previously and had made some fixes in this [PR](https://github.com/OpenLiberty/liberty-language-server/pull/337) around how the workspace is determined for an xml file.

arunvenmany-ibm:
fixes #353

### DIFF
--- a/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/SettingsService.java
+++ b/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/SettingsService.java
@@ -137,7 +137,8 @@ public class SettingsService {
         if (workspace == null) {
             LOGGER.warning("Could not find workspace for server xml URI %s. Variable resolution cannot be performed.".formatted(serverXmlURI));
         } else if (variables != null && variables.containsKey(workspace.getWorkspaceString())) {
-            variableProps = variables.get(workspace.getWorkspaceString());
+            variableProps = variables.get(workspace.getWorkspaceString()) != null ?
+                    variables.get(workspace.getWorkspaceString()) : new Properties();
         } else {
             LOGGER.warning("Could not find variable mapping for workspace URI %s. Variable resolution cannot be performed.".formatted(workspace.getWorkspaceString()));
         }
--- a/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/SettingsService.java
+++ b/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/SettingsService.java
@@ -137,8 +137,7 @@ public class SettingsService {
         if (workspace == null) {
             LOGGER.warning("Could not find workspace for server xml URI %s. Variable resolution cannot be performed.".formatted(serverXmlURI));
         } else if (variables != null && variables.containsKey(workspace.getWorkspaceString())) {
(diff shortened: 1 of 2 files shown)


lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/SettingsService.java
```diff
@@ -137,7 +137,8 @@ public Properties getVariablesForServerXml(String serverXmlURI) {
         if (workspace == null) {
             LOGGER.warning("Could not find workspace for server xml URI %s. Variable resolution cannot be performed.".formatted(serverXmlURI));
         } else if (variables != null && variables.containsKey(workspace.getWorkspaceString())) {
-            variableProps = variables.get(workspace.getWorkspaceString());
+            variableProps = variables.get(workspace.getWorkspaceString()) != null ?
+                    variables.get(workspace.getWorkspaceString()) : new Properties();
```
cherylking, line 141: although this does ensure the Properties returned are never null...we should also figure out why a null Properties has been stored in the Map to begin with...please leave the issue open to address that

cherylking (approved):
The changes here hide the real problem. From an end user perspective, it is nice to not see NPE. But we need to figure out why the Properties in the Map are null for a given server config file.

lemminx-liberty/src/test/java/io/openliberty/LibertyDiagnosticTest.java
```diff
@@ -70,6 +70,7 @@ public void setupWorkspace() {
         libWorkspace = libPM.getLibertyWorkspaceFolders().iterator().next();
         settings= Mockito.mockStatic(SettingsService.class);
         settings.when(SettingsService::getInstance).thenReturn(settingsService);
+        Mockito.lenient().when(settingsService.getVariablesForServerXml(any())).thenReturn(new Properties());
```
arunvenmany-ibm, line 73: @cherylking 
NPE was coming in LibertyDiagnosticeTests because in unit tests we are mocking settingService, since its difficult to create the liberty file structure to populate variables in case unit tests. 
When a service is mocked and any method is not provided when().then() properties. will return null by default


when(settingsService.getVariablesForServerXml(any())).thenReturn(new Properties()) was missing and due to this we were getting NPE.

This is the real line that is resolving the problem now. code changes in settingsService is just made to avoid real life NPE errors

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/services/SettingsService.java
arunvenmany-ibm: Issue is happening in unit tests because we were not mocking the method which returns variableMap, which caused  method to return null. This wont happen in real life scenario as our method is defined to return new Properties() always.

We can see that this is working fine in variable-processing IT

cherylking (approved):

Issue #358, opened 2025-03-20
Diagnostics did not appear in <server description="new server">, even though the dev mode had not been started previously


vaisakhkannan:
I could not see the diagnostics initially, even though I hadn't started dev mode previously. I performed a project clean, then closed the project and reopened the project and server.xml file. Then the diagnostics appeared. I tried this with a fresh project and I found the same issue.

Steps to reproduce

1. Add a varable in server.xml example : `<variable name="truevar" value="99"/>`
2. Use the Variable as the value for one of the element for example : `<httpEndpoint id="defaultHttpEndpoint" httpPort="9080" httpsPort="${truevar}"/>`
3. No diagnostics for the `<server description="new server">`. I haven't started dev mode previously. 

The workaround I used to get the diagnostic is as follows: Perform a project clean, then close and reopen the project and the server.xml file. After that, the diagnostics appeared.

Platform
All OS
liberty-langserver-lemminx-2.2.1
liberty-langserver-2.2.1

arunvenmany-ibm:
Hi @sajeerzeji 

We need to make changes LibertyDiagnosticParticipant#validateVariables.
Current logic assumes that the dev mode is not started if local variable is empty, which wont work due to our recent change to add local variables directly to VariableMap
You need to update the current logic into the following
 If the file contains any <variable> node and dev mode is not started, 
then  show the warning "WARNING: Variable resolution is not available for workspace {0}. Please start the Liberty server for the workspace to enable variable resolution."


To check whether the dev mode is running, you can first check whether liberty-plugin-config.xml file exists in target directory
if  file exists, get <serverDirectory> from the liberty-plugin-config.xml file

I can see that there will be a file created when dev mode running in $serverDirectory/workarea with name ".sRunning"
This file is getting deleted when server is stopped

You can create a utils method for this. There is already a some of the logic present in SettingsService.populateVariablesForWorkspace()

cherylking:
The solution I propose is to just add a check in the current logic to see if the `liberty-plugin-config.xml` file exists (this indicates that the Liberty server has been created which causes the config files to get copied over and makes variable resolution possible). If the file does not exist, the diagnostic should be flagged even if local variables have been identified and added to the `VariableMap`.

PR #362, opened 2025-04-04
Liberty config plugin file validation for proper variables validation

sajeerzeji:
This is a fix for issue #358 

Introduced a new method to check if the liberty-plugin-config.xml is present or not
This check is added at the variables validation to do the diagnostic

### DIFF
--- a/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
+++ b/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
@@ -103,14 +103,13 @@ public class LibertyDiagnosticParticipant implements IDiagnosticsParticipant {
             }
         }
         validateConfigElements(domDocument, diagnosticsList, tempDiagnosticsList, featureGraph, includedFeatures, featureManagerPresent);
-        validateVariables(domDocument,diagnosticsList);
+        validateVariables(domDocument, diagnosticsList, workspace);
     }
 
-    private void validateVariables(DOMDocument domDocument, List<Diagnostic> diagnosticsList) {
+    private void validateVariables(DOMDocument domDocument, List<Diagnostic> diagnosticsList, LibertyWorkspace workspace) {
         String docContent = domDocument.getTextDocument().getText();
         List<VariableLoc> variables = LibertyUtils.getVariablesFromTextContent(docContent);
         Properties variablesMap = SettingsService.getInstance().getVariablesForServerXml(domDocument.getDocumentURI());
-        LibertyWorkspace workspace = LibertyProjectsManager.getInstance().getWorkspaceFolder(domDocument.getDocumentURI());
 
         // Check if the liberty plugin config has been copied to the server or not.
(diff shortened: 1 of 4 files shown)


arunvenmany-ibm (changes requested):
Seems there is some alignment issues with all changes. please reformat the files

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
```diff
@@ -110,8 +110,13 @@ private void validateVariables(DOMDocument domDocument, List<Diagnostic> diagnos
         String docContent = domDocument.getTextDocument().getText();
         List<VariableLoc> variables = LibertyUtils.getVariablesFromTextContent(docContent);
         Properties variablesMap = SettingsService.getInstance().getVariablesForServerXml(domDocument.getDocumentURI());
-        if (variablesMap.isEmpty() && !variables.isEmpty()) {
-            LibertyWorkspace workspace = LibertyProjectsManager.getInstance().getWorkspaceFolder(domDocument.getDocumentURI());
+        LibertyWorkspace workspace = LibertyProjectsManager.getInstance().getWorkspaceFolder(domDocument.getDocumentURI());
```
cherylking, line 113: The workspace was already retrieved in calling method at line 89. Should pass it in to this method.

lemminx-liberty/src/test/java/io/openliberty/LibertyDiagnosticTest.java
```diff
@@ -922,6 +922,7 @@ public void testInvalidVariableDiagnostic() {
         Properties props = new Properties();
         props.putAll(propsMap);
         when(settingsService.getVariablesForServerXml(any())).thenReturn(props);
+        when(settingsService.isLibertyPluginConfigAvailableInServer(any())).thenReturn(true);
```
cherylking, line 925: Are we testing what happens when it returns false anywhere?

cherylking (changes requested):

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
sajeerzeji: Changed this part like it will accept the `LibertyWorkspace` object as an argument in the method `validateVariables` and from `validateDom` we are passing the `LibertyWorkspace` to `validateVariables`.

lemminx-liberty/src/test/java/io/openliberty/LibertyDiagnosticTest.java
sajeerzeji: Added a new diagnostics unit test for no configuration file available scenario. For this, we will set `when(settingsService.isLibertyPluginConfigAvailableInServer(any())).thenReturn(false);` there.
Also moved the `when(settingsService.isLibertyPluginConfigAvailableInServer(any())).thenReturn(true);` to the `@BeforeEach` method so that we don't need to have it in each tests.

cherylking (approved):

arunvenmany-ibm (approved):

Issue #70, opened 2022-06-10
Flag incompatible features in server.xml


cherylking:
When a user adds a feature to the server.xml, dev mode will automatically install it. If auto feature generation is enabled, it will also detect incompatible features. The language server should use that information from the Maven/Gradle plugin to flag any incompatible features that have been configured. Determine if there is any other way the language server could determine incompatible features for flagging.

This is a future enhancement that is not part of the MVP. Slide 22 of UFO.

scottkurz:
So one idea here would be to call the Liberty runtime install features function, like the `install-feature` goal does today presumably.

Is it worth considering as an alternative an idea more like:  "Look at the (cached, maximal) featurelist XML" and build a graph of "enables" (and "enabled by") relationships.   In particular look for which of the umbrella features:  "jakartaee-10.0", "microProfile-5.0" enable a given feature.   Give a visual indicator (on hover?) which aggregate a given API feature is enabled by and give a warning if there seem to be conflicts (e.g. an EE 9 feature and an EE 10 feature).

Just looking at this should raise several questions... is it good to have a "shorthand" of "this is an EE 9 feature" when the runtime doesn't quite say exactly that.  (What if a given feature like batch-2.1 doesn't have an update in EE 11 so it's part of both?).

What about OL vs WL?   Does trying to use a cached feature list introduce all the issues of the cached XSD discussion (e..g beta vs ga as discussed here: https://github.com/OpenLiberty/liberty-language-server/issues/259#issuecomment-1889551015  ).

Perhaps there is no better way than calling the runtime... 

If not, though, I do wonder if there is still value in trying to display a shorthand EE platform version indicator (e.g. "EE10") of some form.  Maybe it'd need to go in the metadata to get gen'd into the XSD.   OK, that's enough for this comment.

scottkurz:
UDPATE: DXDI 2024-01-17

One key point to understand is that with the completion of the generateFeatures function, to the point that it can be enabled by default, there will be less need for this type of thing.  

There is a thought that we could invoke the binary scanner to help flag incompatible feature combinations, in a way that would be cheaper than the invocation of the Liberty runtime which we call when doing the liberty-maven-plugin 'install-feature' goal.  It seems though that such an effort would be best undertaken after the next (final) stage of the generateFeatures work is complete.

In the meantime, it seems like we could take two shorter-term actions 

[...]

PR #364, opened 2025-04-15
Features incompatibility diagnostics for versioned features in server.xml and added a new unit test

sajeerzeji:
This is a fix for #70 

A diagnostics for incompatible versioned feature in server.xml is added.
This will check if the features have any common platforms, hence consider those as compatible features, otherwise will show error diagnostics.
A new unit test also added for this compatibility check.


lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
```diff
@@ -679,4 +679,128 @@ private static void validateVariableExists(DOMDocument domDocument, List<Diagnos
...
+        // Get all features and their nodes from the featureManager
+        Map<String, DOMNode> featureNodes = LibertyUtils.getAllFeatureNodes(featureManagerNode);
+        List<String> configuredFeatures = new ArrayList<>(featureNodes.keySet());
+
+        if (configuredFeatures.size() <= 1) {
+            return; // Need at least two features to check compatibility
+        }
+
+        // Skip if already there are some diagnostics (This is implemented to ensure that some of the Unit Tests are not broken).
+        if (diagnosticsList != null && !diagnosticsList.isEmpty()) {
```
arunvenmany-ibm, line 702: as discussed, this check cannot be here, as this will skip incompatible check if there is any duplicate feature. 
For each feature with incompatibility , check whether a diagnostic exists with same range

arunvenmany-ibm (changes requested):
@cherylking  need your input on this, 

PR is considering javaee and jakartaee as separate platforms and grouping features of each platform separately and finding common platforms. 
Due to this, if we use "jsp-2.3" and "pages-4.0" together, it will not throw any incompatible feature error because jsp uses javaee platform and pages use jakartaee platform . Should we do a cross platform check for javaee and jakartaee?

The current PR is implemented to show feature incompatibility daignostics only if there is no duplicate features(servlet 3.1 and servlet 4.0) or no same feature with different name( like jsp and pages). Due to this we will see the duplicate version diagnostics

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
sajeerzeji: This part have fixed, the check have been moved to line level, so it will now check if the line have any diagnostics present or not and will add this diagnostic if only there is no other diagnostics for the line.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
```diff
@@ -35,14 +35,7 @@
...
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Properties;
-import java.util.Set;
+import java.util.*;
```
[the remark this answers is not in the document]
sajeerzeji: I have removed the wild card imports and enabled as suggested in my IDE settings.

cherylking:
> The current PR is implemented to show feature incompatibility daignostics only if there is no duplicate features(servlet 3.1 and servlet 4.0) or no same feature with different name( like jsp and pages).

Why would we not want to show the incompatible feature message in this scenario?

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
cherylking: Is it not possible to have more than one diagnostic on a particular line?

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
```diff
@@ -679,4 +687,122 @@ private static void validateVariableExists(DOMDocument domDocument, List<Diagnos
...
+        // Find the featureManager element to get feature nodes if available
+        DOMNode featureManagerNode = LibertyUtils.getFeatureManagerElement(domDocument);
+        Map<String, DOMNode> featureNodes = featureManagerNode != null ?
+                LibertyUtils.getAllFeatureNodes(featureManagerNode) :
+                new HashMap<>();
+
+        // Categorize features by platform type
+        for (String feature : versionedFeatures) {
+            Set<String> platforms = FeatureService.getInstance().getAllPlatformsForFeature(feature, libertyVersion, libertyRuntime, requestDelay, domDocument.getDocumentURI());
+            DOMNode featureNode = featureNodes.getOrDefault(feature, null);
```
cherylking, line 723: versionedFeatures has feature names in all lower case. The featureNodes Map I believe has the original case for the features as they appeared in the file. This means you can get null here even though the feature is present. I am not sure why you have this Map when you could have passed in the feature nodes and just iterated through them.

cherylking (changes requested):

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
```diff
@@ -679,4 +687,122 @@ private static void validateVariableExists(DOMDocument domDocument, List<Diagnos
...
+    private void validateFeatureCompatibility(Set<String> versionedFeatures, DOMDocument domDocument, List<Diagnostic> diagnosticsList) {
+        if (versionedFeatures == null || versionedFeatures.size() <= 1) {
+            return; // Need at least two features to check compatibility
+        }
+
+        // Get Liberty version and runtime from the document
+        LibertyRuntime runtimeInfo = LibertyUtils.getLibertyRuntimeInfo(domDocument);
+        String libertyVersion = runtimeInfo == null ? null : runtimeInfo.getRuntimeVersion();
+        String libertyRuntime = runtimeInfo == null ? null : runtimeInfo.getRuntimeType();
+        int requestDelay = SettingsService.getInstance().getRequestDelay();
```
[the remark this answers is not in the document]
sajeerzeji: `libertyVersion`, `libertyRuntime` and `requestDelay` are passed in from the calling method.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
```diff
@@ -679,4 +687,122 @@ private static void validateVariableExists(DOMDocument domDocument, List<Diagnos
...
+
+        // Maps to track features by platform type
+        Map<String, Map<String, Object>> javaEEFeatures = new HashMap<>();
+        Map<String, Map<String, Object>> jakartaEEFeatures = new HashMap<>();
```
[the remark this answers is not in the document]
sajeerzeji: `javaEEFeatures` and `jakartaEEFeatures` are combined into `eeFeatures`

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
```diff
@@ -679,4 +687,122 @@ private static void validateVariableExists(DOMDocument domDocument, List<Diagnos
...
+        Map<String, Map<String, Object>> microProfileFeatures = new HashMap<>();
+
+        // Find the featureManager element to get feature nodes if available
+        DOMNode featureManagerNode = LibertyUtils.getFeatureManagerElement(domDocument);
+        Map<String, DOMNode> featureNodes = featureManagerNode != null ?
```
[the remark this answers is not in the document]
sajeerzeji: `featureNodes` is now passed from the calling method, hence we don't need the `featureManagerNode` as it was used to fetch `featureNodes`

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyUtils.java
```diff
@@ -692,4 +695,73 @@ public static void checkAndAddNewVariables(DOMDocument document, Properties vari
...
+            return featureNodes;
+        }
+
+        List<DOMNode> features = featureManagerNode.getChildren();
+        for (DOMNode featureNode : features) {
+            if ("feature".equals(featureNode.getLocalName())) {
+                DOMNode featureTextNode = (DOMNode) featureNode.getChildNodes().item(0);
+                String featureName = featureTextNode.getTextContent();
+                if (featureName != null && !featureName.trim().isEmpty()) {
+                    featureNodes.put(featureName.trim(), featureNode);
```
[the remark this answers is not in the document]
sajeerzeji: This was orginal featureName that presevers the case. But this method is removed as we are passing the `featureNodes` from the calling method to `validateFeatureCompatibility` now. This method was indented to fetch feature nodes from featureManager, and it is no longer needed in new changes.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
sajeerzeji: We can display multiple diagnostics here. However, is it beneficial to do so?
We are currently limiting showing the incompatibility diagnostic only to "duplicate features" (same feature with same/different versions) and "name changed features" (e.g., jsp-2.3 & pages-3.0). In these specific cases, is it good to indicate their incompatibility?

sajeerzeji:
> This is a fix for #70
> 
> A diagnostics for incompatible versioned feature in server.xml is added. This will check if the features have any common platforms, hence consider those as compatible features, otherwise will show error diagnostics. A new unit test also added for this compatibility check.

The sample diagnostic will look like this
image

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
cherylking: so in both of those cases...they should remove one of the features...after doing so we may show an incompatible feature diagnostic...but they may need that info in advance...especially for the one where the name changed.

I think we should show both messages. Can you add an image here of what that would look like?

sajeerzeji: It will show both messages like this
image

sajeerzeji: I have changed this part to show both diagnostics together. This includes some changes in some of the existing unit test cases as well.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
```diff
@@ -790,18 +789,18 @@ private void validatePlatformTypeCompatibility(Map<String, Map<String, Object>>
             for (Map.Entry<String, Map<String, Object>> entry : featureMap.entrySet()) {
                 DOMNode featureNode = (DOMNode) entry.getValue().get("node");
                 if (featureNode == null) continue;
+                DOMNode featureTextNode = (DOMNode) featureNode.getChildNodes().item(0);
+                if (featureTextNode == null) continue;
```
cherylking, line 793: Why would this be null?

lemminx-liberty/src/main/resources/messages/libertyBundles.properties
```diff
@@ -29,6 +29,7 @@ liberty.lemminx.versionless.feature.no.supported.platform.error=ERROR: The \"{0}
 liberty.lemminx.variable.name.invalid.error=ERROR: The variable should have a valid name defined in name attribute.
 liberty.lemminx.variable.value.attribute.invalid.error=ERROR: The variable \"{0}\" should have value or defaultValue attribute defined.
 liberty.lemminx.variable.does.not.exist.error=ERROR: The variable \"{0}\" does not exist.
+liberty.lemminx.features.incompatible.error=ERROR: The feature {0} is incompatible with {1}. They are not sharing any common platforms.
```
cherylking, line 32: Change the second sentence - `The features do not share a common platform.`

cherylking (changes requested):

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
```diff
@@ -728,15 +734,9 @@ private void validateFeatureCompatibility(Set<String> versionedFeatures, DOMDocu
...
-                if (platform.toLowerCase().startsWith("javaee-")) {
-                    Map<String, Object> featureData = javaEEFeatures.computeIfAbsent(feature, k -> new HashMap<>());
-                    featureData.computeIfAbsent("features", k -> new HashSet<String>());
-                    ((Set<String>) featureData.get("features")).add(platform);
-                    if (featureNode != null) {
-                        featureData.put("node", featureNode);
-                    }
-                } else if (platform.toLowerCase().startsWith("jakartaee-")) {
-                    Map<String, Object> featureData = jakartaEEFeatures.computeIfAbsent(feature, k -> new HashMap<>());
+                if (platform.toLowerCase().startsWith("javaee-") || platform.toLowerCase().startsWith("jakartaee-")) {
```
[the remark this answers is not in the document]
sajeerzeji: I have changed this code as it was complex and confusing. It is now categorizing the features into sets of Strings. I kept the logic of fetching feature nodes, which is by making a map, because I feel it is faster this way to fetch feature nodes by feature name.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
```diff
@@ -790,18 +789,18 @@ private void validatePlatformTypeCompatibility(Map<String, Map<String, Object>>
             for (Map.Entry<String, Map<String, Object>> entry : featureMap.entrySet()) {
                 DOMNode featureNode = (DOMNode) entry.getValue().get("node");
                 if (featureNode == null) continue;
```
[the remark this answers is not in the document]
sajeerzeji: Removed this redundant null check.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyDiagnosticParticipant.java
sajeerzeji: Removed this redundant null check.

lemminx-liberty/src/main/resources/messages/libertyBundles.properties
sajeerzeji: Changed the second sentence of diagnostic message to `The features do not share a common platform.`

cherylking (approved):
The new code is much easier to understand. Thanks for reworking it!

arunvenmany-ibm (approved):

Issue #383, opened 2025-08-06
SVT: Invalid locale format: en-us error encountered using preGA 25.0.8 VSCode


hanczaryk:
SVT: Invalid locale format: en-us error encountered using preGA 25.0.8 VSCode

I've installed the latest preGA 25.0.8 VSCode extension on my Windows laptop.

After opening VSCode, the Output's XML Support shows the following Error message.
```
[Info  - 11:04:29 AM] Aug 06, 2025 11:04:29 org.eclipse.lemminx.XMLLanguageServer initialize()
Message: Initializing XML Language server
LemMinX Server info:
 - Version : 0.31.0
 - Java : C:\java21
 - VM Version : 21.0.2
 - Git de7f529 - [maven-release-plugin] prepare release 0.31.0
[Info  - 11:04:29 AM] Aug 06, 2025 11:04:29 io.openliberty.tools.langserver.lemminx.services.LibertyProjectsManager setWorkspaceFolders()
Message: Checking Liberty workspace for sub-modules: file:/c%3A/Users/XXX/git/MPProShop
[Info  - 11:04:29 AM] Aug 06, 2025 11:04:29 io.openliberty.tools.langserver.lemminx.services.LibertyProjectsManager setWorkspaceFolders()
Message: Skipping parent module: file:/c:/Users/XXX/git/MPProShop/
[Info  - 11:04:29 AM] Aug 06, 2025 11:04:29 io.openliberty.tools.langserver.lemminx.services.LibertyProjectsManager setWorkspaceFolders()
Message: Adding Liberty workspace by default: file:/c%3A/Users/XXX/git/MPProShop
[Error - 11:04:29 AM] Aug 06, 2025 11:04:29 org.eclipse.lemminx.services.extensions.XMLExtensionsRegistry registerExtension()
Message: Error while initializing extension <io.openliberty.tools.langserver.lemminx.LibertyExtension>
java.lang.IllegalArgumentException: Invalid locale format: en-us
    at org.apache.commons.lang3.LocaleUtils.toLocale(LocaleUtils.java:133)
    at io.openliberty.tools.langserver.lemminx.services.SettingsService.initializeLocale(SettingsService.java:160)
    at io.openliberty.tools.langserver.lemminx.LibertyExtension.start(LibertyExtension.java:60)
    at org.eclipse.lemminx.services.extensions.XMLExtensionsRegistry.registerExtension(XMLExtensionsRegistry.java:284)
[...]

PR #384, opened 2025-08-14
introduced locale fallback methods

arunvenmany-ibm:
resolving #383 

Added code changes to find locale from any locale formats(en-us,en_US,en)
Also added code to match the current locale with existing liberty allowed locales and default to english if no match is found

Tested in vscode using different locales set using argv.json and directly by configuring language
Language server now accepts valid locales in all standard formats and if it cannot be matched, locale is defaulted to english

Test Video size is huge, uploading to [onedrive](https://ibm-my.sharepoint.com/:v:/p/arun_kumar_v_n/EcJuhdyCpptEkXRHQrZB_CkBXe8coDLCkJUpaMEk6U1O_w?nav=eyJyZWZlcnJhbEluZm8iOnsicmVmZXJyYWxBcHAiOiJPbmVEcml2ZUZvckJ1c2luZXNzIiwicmVmZXJyYWxBcHBQbGF0Zm9ybSI6IldlYiIsInJlZmVycmFsTW9kZSI6InZpZXciLCJyZWZlcnJhbFZpZXciOiJNeUZpbGVzTGlua0NvcHkifX0&e=w9g7WI)

### DIFF
--- a/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/ResourceBundleUtil.java
+++ b/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/ResourceBundleUtil.java
@@ -92,8 +92,8 @@ public final class ResourceBundleUtil {
             return new Locale(parts[0], parts[1], parts[2]);
         }
         // If all else fails, return a default locale
-        LOGGER.warning("Setting locale as en_US as initializeParams locale is null");
-        return Locale.US;
+        LOGGER.warning("Setting locale as en as initializeParams locale is null");
+        return Locale.ENGLISH;
     }
 
     /**
--- a/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/ResourceBundleUtil.java
+++ b/lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/ResourceBundleUtil.java
@@ -13,8 +13,12 @@
 package io.openliberty.tools.langserver.lemminx.util;
 
 import io.openliberty.tools.langserver.lemminx.services.SettingsService;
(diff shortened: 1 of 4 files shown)


lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/ResourceBundleUtil.java
```diff
@@ -55,4 +59,82 @@ public static String getMessage(String key, Object... args) {
...
+                return availableLocale;
+            }
+        }
+        // Check for a language-only match
+        for (Locale availableLocale : getAvailableLocales()) {
+            if (availableLocale.getLanguage().equals(userLocale.getLanguage())) {
+                return availableLocale;
+            }
+        }
+        // Check for a "parent" locale match (e.g., "en" for "en-US")
```
cherylking, line 132: Which one of these `if` statements is going to handle the last two characters in lowercase when they should be uppercase? So if `en-us` is passed in, what code is recognizing that?

cherylking: We need to make sure we will not return `zh_CN` for `zh_TW` for example if `zh-tw` is passed in. I don't see any code logic that is uppercasing the last two characters and checking.

arunvenmany-ibm: Locale.forLanguageTag() method takes care of the locale in different cases
https://docs.oracle.com/javase/8/docs/api/java/util/Locale.html#forLanguageTag-java.lang.String-

Also added tests for zh_CN and zh_TW

PR #386, opened 2025-08-28
 changes to download latest xsd schema

venmanyarun:
1. Fetches latest maven-metadata.xml from https://repo1.maven.org/maven2/io/openliberty/features/open_liberty_featurelist/maven-metadata.xml
Since latest version is 25.0.0.8, latest version xsd is downloaded
image
image

2. Fetches xsd with locale
image
image

3. Using liberty runtime version 25.0.0.7 and libertyDev, xsd is changed to ol-25.0.0.7.xsd, which is generated
image

4. Fallback method, deleted .lemminx directory and disconnected internet, should go to cached 25.0.0.6 , which is cached in classpath, copied to .lemminx folder

image
image

### DIFF
--- a/lemminx-liberty/src/test/java/io/openliberty/LibertyHoverTest.java
+++ b/lemminx-liberty/src/test/java/io/openliberty/LibertyHoverTest.java
@@ -3,6 +3,7 @@ package io.openliberty;
 import io.openliberty.tools.langserver.lemminx.services.LibertyProjectsManager;
 import io.openliberty.tools.langserver.lemminx.services.LibertyWorkspace;
 import io.openliberty.tools.langserver.lemminx.services.SettingsService;
+import org.apache.commons.io.FilenameUtils;
 import org.eclipse.lemminx.uriresolver.CacheResourcesManager;
 import org.eclipse.lsp4j.WorkspaceFolder;
 import org.junit.jupiter.api.AfterEach;
@@ -15,7 +16,6 @@ import static io.openliberty.tools.langserver.lemminx.LibertyXSDURIResolver.LIBE
 
 import org.eclipse.lemminx.XMLAssert;
 import org.eclipse.lemminx.commons.BadLocationException;
-import org.junit.jupiter.api.condition.DisabledOnOs;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.Mock;
 import org.mockito.MockedStatic;
@@ -26,12 +26,12 @@ import static io.openliberty.tools.langserver.lemminx.util.LibertyConstants.BRAZ
 import static io.openliberty.tools.langserver.lemminx.util.LibertyConstants.LOCALE;
 import static io.openliberty.tools.langserver.lemminx.util.LibertyConstants.VERSION;
 import static org.eclipse.lemminx.XMLAssert.r;
-import static org.junit.jupiter.api.condition.OS.WINDOWS;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.when;
 
 import java.io.File;
 import java.io.IOException;
+import java.net.URL;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -42,6 +42,7 @@ import java.util.Properties;
 @ExtendWith(MockitoExtension.class)
 public class LibertyHoverTest {
 
+        public static final String TEST_VERSION = "25.0.0.8";
         @Mock
         SettingsService settingsService;
 
@@ -223,9 +224,9 @@ public class LibertyHoverTest {
 
         @Test
         public void testXSDSchemaHoverWithLatestVersion() throws BadLocationException, IOException {
-                when(settingsService.getLatestRuntimeVersion()).thenReturn("25.0.0.8");
+                when(settingsService.getLatestRuntimeVersion()).thenReturn(TEST_VERSION);
                 String serverXSDURI = new CacheResourcesManager.ResourceToDeploy(
-                        LIBERTY_SCHEMA_VERSION_XSD.replace(VERSION, "25.0.0.8"), LIBERTY_SCHEMA_VERSION_XSD)
+                        LIBERTY_SCHEMA_VERSION_XSD.replace(VERSION, TEST_VERSION), LIBERTY_SCHEMA_VERSION_XSD)
                         .getDeployedPath().toUri().toString().replace("///", "/");
 
                 String serverXML = String.join(newLine, //
@@ -238,16 +239,16 @@ public class LibertyHoverTest {
 
                 XMLAssert.assertHover(serverXML, serverXMLURI, "Defines how the server loads features." + //
                                 System.lineSeparator() + System.lineSeparator() + //
-                                "Source: [open_liberty_schema-25.0.0.8.xsd](" + serverXSDURI + ")", //
+                                "Source: ["+ FilenameUtils.getName(serverXSDURI) +"](" + serverXSDURI + ")", //
                         r(1, 8, 1, 22));
         }
 
         @Test
         public void testXSDSchemaHoverWithInvalidLocale() throws BadLocationException, IOException {
-                when(settingsService.getLatestRuntimeVersion()).thenReturn("25.0.0.8");
+                when(settingsService.getLatestRuntimeVersion()).thenReturn(TEST_VERSION);
                 when(settingsService.getCurrentLocale()).thenReturn(new Locale("hi","ind"));
                 String serverXSDURI = new CacheResourcesManager.ResourceToDeploy(
-                        LIBERTY_SCHEMA_VERSION_XSD.replace(VERSION, "25.0.0.8"), LIBERTY_SCHEMA_VERSION_XSD)
+                        LIBERTY_SCHEMA_VERSION_XSD.replace(VERSION, TEST_VERSION), LIBERTY_SCHEMA_VERSION_XSD)
                         .getDeployedPath().toUri().toString().replace("///", "/");
                 String serverXML = String.join(newLine, //
                         "<server description=\"Sample Liberty server\">", //
@@ -256,19 +257,19 @@ public class LibertyHoverTest {
                         "       </featureManager>", //
                         "</server>" //
                 );
-                // should default back to 25.0.0.8
+                // should default back to TEST_VERSION
                 XMLAssert.assertHover(serverXML, serverXMLURI, "Defines how the server loads features." + //
                                 System.lineSeparator() + System.lineSeparator() + //
-                                "Source: [open_liberty_schema-25.0.0.8.xsd](" + serverXSDURI + ")", //
+                                "Source: ["+ FilenameUtils.getName(serverXSDURI) +"](" + serverXSDURI + ")", //
                         r(1, 8, 1, 22));
         }
 
         @Test
         public void testXSDSchemaHoverWithValidLocale() throws BadLocationException, IOException {
-                when(settingsService.getLatestRuntimeVersion()).thenReturn("25.0.0.8");
+                when(settingsService.getLatestRuntimeVersion()).thenReturn(TEST_VERSION);
                 when(settingsService.getCurrentLocale()).thenReturn(BRAZIL_PORTUGESE_LOCALE);
                 String serverXSDURI = new CacheResourcesManager.ResourceToDeploy(
-                        LIBERTY_SCHEMA_VERSION_WITH_LOCALE_XSD.replace(VERSION, "25.0.0.8")
+                        LIBERTY_SCHEMA_VERSION_WITH_LOCALE_XSD.replace(VERSION, TEST_VERSION)
                                 .replace(LOCALE, "pt_BR"), LIBERTY_SCHEMA_VERSION_WITH_LOCALE_XSD)
                         .getDeployedPath().toUri().toString().replace("///", "/");
                 String serverXML = String.join(newLine, //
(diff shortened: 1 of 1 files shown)


lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyVersionDownloadUtil.java
```diff
@@ -0,0 +1,274 @@
...
+    /**
+     * fetches a resource from remote repository
+     * if fetch is successful and cache path is present, file will be written to cache path as well
+     *
+     * @param resourceUrl   resource url
+     * @param cacheFilePath cache location
+     * @return resource content in string
+     */
+    public static String getResource(String resourceUrl, String cacheFilePath) {
+        HttpClient client = HttpClient.newBuilder()
```
cherylking, line 91: I am curious why you chose to use HttpClient and HttpRequest for this? It seems like you could just construct the URL object, open the stream and copy the file locally.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyVersionDownloadUtil.java
```diff
@@ -0,0 +1,274 @@
...
+        if (SettingsService.getInstance().getCurrentLocale() == null || SettingsService.getInstance().getCurrentLocale().equals(Locale.US)
+                || SettingsService.getInstance().getCurrentLocale().equals(Locale.ENGLISH) || urlWithLocale == null) {
+            schemaURL = url.replace(VERSION, rtVersion);
+        } else {
+            Locale currentLocale = SettingsService.getInstance().getCurrentLocale();
+            // liberty published file names contain only language code as locale for all locales except Portuguese (Brazil) and Traditional Chinese
+            // filename samples
+            // https://repo1.maven.org/maven2/io/openliberty/features/open_liberty_schema_es/25.0.0.8/open_liberty_schema_es-25.0.0.8.xsd
+            // https://repo1.maven.org/maven2/io/openliberty/features/open_liberty_schema_pt_BR/25.0.0.8/open_liberty_schema_pt_BR-25.0.0.8.xsd
+            schemaURL = urlWithLocale.replace(VERSION, rtVersion).replace("$LOCALE", currentLocale.getLanguage());
```
cherylking, line 198: It is confusing to have VERSION replaced with a variable that is actually `$VERSION` and LOCALE does not have a variable. So at first glance I was trying to figure out why we don't have to replace `$VERSION` the same way we replace `$LOCALE`. Can we be consistent?

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyVersionDownloadUtil.java
```diff
@@ -0,0 +1,274 @@
...
+            if (Locale.TAIWAN.equals(currentLocale) || new Locale("pt", "BR").equals(currentLocale)) {
```
cherylking, line 199: Let's create a final static variable for the pt_BR Locale that we can use in this if statement. No need to create it every time.

lemminx-liberty/src/test/java/io/openliberty/LibertyHoverTest.java
```diff
@@ -210,4 +213,45 @@ public void testVariableHover() throws BadLocationException {
...
+                String serverXSDURI = SERVER_XSD_RESOURCE_DEFAULT.getDeployedPath().toUri().toString().replace("///", "/");
+
+                String serverXML = String.join(newLine, //
+                        "<server description=\"Sample Liberty server\">", //
+                        "       <feature|Manager>", //
+                        "               <feature>jaxrs-2.1</feature>", //
+                        "       </featureManager>", //
+                        "</server>" //
+                );
+                // should default back to 25.0.0.6
```
cherylking, line 251: Since the latest version is 25.0.0.8 in this test case method, should we consider falling back to the English schema for 25.0.0.8 vs the default cached English schema which is 25.0.0.6? Basically, I think we should detect that the locale is not supported before we even try to download the latest schema for it, and instead download the latest schema in English and use it.

cherylking (changes requested):

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyVersionDownloadUtil.java
arunvenmany-ibm: HttpClient is the standardized API which is non blocking, compared to java.net.URL.openStream()

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyVersionDownloadUtil.java
```diff
@@ -0,0 +1,274 @@
...
+
+    private LibertyVersionDownloadUtil() {
+    }
+
+    // The URL for the maven-metadata.xml file.
+    private static String METADATA_URL = "https://repo1.maven.org/maven2/io/openliberty/features/open_liberty_featurelist/maven-metadata.xml";
+
+    // The path to the local cache file, following the specified convention.
+    private static String CACHE_BASE_DIR = System.getProperty("user.home") + File.separator + ".lemminx" + File.separator + "cache";
+    private static String CACHE_FILE_PATH = CACHE_BASE_DIR + File.separator + "https" + File.separator + "repo1.maven.org" + File.separator + "maven2" + File.separator + "io" + File.separator + "openliberty" + File.separator + "features" + File.separator + "open_liberty_featurelist" + File.separator + "maven-metadata.xml";
```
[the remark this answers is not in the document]
arunvenmany-ibm: changing to use java.nio.path

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyVersionDownloadUtil.java
arunvenmany-ibm: Done

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyExtension.java
```diff
@@ -58,6 +61,13 @@ public void start(InitializeParams initializeParams, XMLExtensionsRegistry xmlEx
             LOGGER.warning("Could not get workspace folders: " + e.toString());
         }
         SettingsService.getInstance().initializeLocale(initializeParams);
+        String latestVersion = LibertyVersionDownloadUtil.getLatestVersionFromMetadata();
+        if (latestVersion != null) {
+            LOGGER.info("Latest Open Liberty version found: %s".formatted(latestVersion));
```
cherylking, line 66: I don't think this should be an info message...either debug or finer.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyVersionDownloadUtil.java
```diff
@@ -0,0 +1,292 @@
...
+            urlWithLocale == null || 
+            Locale.US.equals(currentLocale) || 
+            Locale.ENGLISH.equals(currentLocale) || 
+            findBestMatchingLocale(currentLocale) == null) {
+            return url.replace(VERSION, rtVersion != null ? rtVersion : DEFAULT_LIBERTY_VERSION);
+        }
+        
+        // Special handling for specific locales that need full locale code
+        if (Locale.TAIWAN.equals(currentLocale) || BRAZIL_PORTUGESE_LOCALE.equals(currentLocale)) {
+            return urlWithLocale.replace(VERSION, rtVersion)
```
cherylking, line 213: Could `rtVersion` be null here and in the following return statement at line 218? At line 208 you check for `rtVersion != null` and fall back to DEFAULT_LIBERTY_VERSION if it is null.

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyVersionDownloadUtil.java
```diff
@@ -0,0 +1,292 @@
...
+
+    /**
+     * @param url           schema or featurelist url
+     * @param urlWithLocale url with locale placeholder
+     * @return downloaded resource path
+     */
+    public static Path downloadAndCacheLatestResource(String url, String urlWithLocale) {
+        Path serverResourceFile = null;
+        if (SettingsService.getInstance().getLatestRuntimeVersion() != null) {
+            String resourceURL = parseURL(url, urlWithLocale);
```
cherylking, line 230: So it looks like you cannot get into method `parseURL` with a null latest runtime version? Part of that method was checking for null rtVersion.

lemminx-liberty/src/test/java/io/openliberty/LibertyHoverTest.java
cherylking: change comment at line 258 to match what the test is verifying

lemminx-liberty/src/test/java/io/openliberty/LibertyHoverTest.java
```diff
@@ -210,4 +219,71 @@ public void testVariableHover() throws BadLocationException {
...
+                XMLAssert.assertHover(serverXML, serverXMLURI, "Defines how the server loads features." + //
+                                System.lineSeparator() + System.lineSeparator() + //
+                                "Source: [open_liberty_schema-25.0.0.8.xsd](" + serverXSDURI + ")", //
+                        r(1, 8, 1, 22));
+        }
+
+        // this test is disabled for windows as hover description in french has special characters, which fails assert for description
```
cherylking, line 265: as we discussed, try to find a different hover description without accents/special characters so that this test can run on Windows also.

cherylking (changes requested):

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyExtension.java
```diff
@@ -58,6 +59,13 @@ public void start(InitializeParams initializeParams, XMLExtensionsRegistry xmlEx
...
+            SettingsService.getInstance().setLatestRuntimeVersion(latestVersion);
+        } else {
+            LOGGER.warning("Could not determine the latest Open Liberty version. Using default version");
```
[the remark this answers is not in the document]
arunvenmany-ibm: moved default version to constants file and referred in LibertyXSDURIResolver

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyVersionDownloadUtil.java
arunvenmany-ibm: Yes.. to be on the safer side, I have added additional checks

lemminx-liberty/src/test/java/io/openliberty/LibertyHoverTest.java
arunvenmany-ibm: Done.. pt_BR works

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/util/LibertyVersionDownloadUtil.java
arunvenmany-ibm: Done

lemminx-liberty/src/main/java/io/openliberty/tools/langserver/lemminx/LibertyExtension.java
arunvenmany-ibm: Done

lemminx-liberty/src/test/java/io/openliberty/LibertyHoverTest.java
arunvenmany-ibm: Done

lemminx-liberty/src/test/java/io/openliberty/LibertyHoverTest.java
```diff
@@ -255,23 +256,20 @@ public void testXSDSchemaHoverWithInvalidLocale() throws BadLocationException, I
                         "       </featureManager>", //
                         "</server>" //
                 );
-                // should default back to 25.0.0.6
+                // should default back to 25.0.0.8
                 XMLAssert.assertHover(serverXML, serverXMLURI, "Defines how the server loads features." + //
                                 System.lineSeparator() + System.lineSeparator() + //
                                 "Source: [open_liberty_schema-25.0.0.8.xsd](" + serverXSDURI + ")", //
```
cherylking, line 262: Had one more thought....you could use the constant for the default version here so that the test case will not need to be updated...it would always verify that the default is used.

lemminx-liberty/src/test/java/io/openliberty/LibertyHoverTest.java
```diff
@@ -280,10 +278,10 @@ public void testXSDSchemaHoverWithValidLocale() throws BadLocationException, IOE
                         "       </featureManager>", //
                         "</server>" //
                 );
-                // should default back to 25.0.0.6
-                XMLAssert.assertHover(serverXML, serverXMLURI, "Définit comment le serveur charge les fonctions." + //
+                // should default back to 25.0.0.8
+                XMLAssert.assertHover(serverXML, serverXMLURI, "Define como o servidor carrega recursos." + //
                                 System.lineSeparator() + System.lineSeparator() + //
-                                "Source: [open_liberty_schema_fr-25.0.0.8.xsd](" + serverXSDURI + ")", //
+                                "Source: [open_liberty_schema_pt_BR-25.0.0.8.xsd](" + serverXSDURI + ")", //
```
cherylking, line 284: same comment here about using constant for default version

cherylking (changes requested):

lemminx-liberty/src/test/java/io/openliberty/LibertyHoverTest.java
cherylking: Sorry, but I mistook this test case to be testing the default (which is actually 25.0.0.6). Existing test cases that are actually testing the default could use the DEFAULT_LIBERTY_VERSION from LibertyConstants. This test is just defaulting to the latest which will keep changing. Not much you can do there.

cherylking (approved):

